PR #1, opened 2022-03-02
[MGDOBR-344] Initial application setup

kelvah:
Initial setup of the bridge sandbox UI as a federated module.

PR #3, opened 2022-03-04
[MGDOBR-375] Setup internal routing

kelvah:
Follows #1


manstis (approved):
I can guess the use of `useBasename()`.

It will give us the base URL for when running embedded in App Services?

kelvah:
> I can guess the use of `useBasename()`.
> 
> It will give us the base URL for when running embedded in App Services?

@manstis That's correct. It's a function shared via a `BasenameContext` inside App Services. It's located [here](https://github.com/redhat-developer/app-services-ui/blob/main/src/app/utils/getBaseName.ts).

PR #8, opened 2022-03-21
    MGDOBR-445: Update generated SDK due to openapi changes

jomarko:
This PR is built on top of #6 and is dependent on changes in 5733d9e2be6485d52ffa08870cabdee0/sandbox#436

Due to the dependencies, this PR is out of date, demonstrates just "kind of changes", but to see final changes we need to:
- merge 5733d9e2be6485d52ffa08870cabdee0/sandbox#436
- merge #6

    
For more details see:
- https://issues.redhat.com/browse/MGDOBR-445


scripts/fetch_openapi.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# get the download URL of the OpenAPI spec file
+#OPENAPI_FILE_URL="https://raw.githubusercontent.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/main/openapi/openapi.yaml"
+OPENAPI_FILE_URL="https://raw.githubusercontent.com/jomarko/sandbox/MGDOBR-445/openapi/openapi.yaml"
+OPENAPI_FILE_NAME=$(dirname $0)/../openapi/smart-events.yaml
```
jomarko, line 6: needs to be reverted once reviewed

scripts/generate_openapi_client.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# remove old generated models
+
+rm -Rf $output_path/model $output_path/api
+
+# generate an API client for smart-events
+
+echo "Generating source code based on ${openapi_filename}"
+
+npx @openapitools/openapi-generator-cli generate -g typescript-axios --api-name-suffix "sdk" -i \
```
jomarko, line 26: `--api-name-suffix "sdk"` is to have file names `bridges-apisdk.ts` instead of `bridges-apiapi.ts`

jomarko:
@kelvah @vpellegrino @manstis I think this is now worth to review. Please notice quite a lot very similar model files named like `inline-response2XXX.ps`. That seems to me as worth to investigate in future. I reported https://issues.redhat.com/browse/MGDOBR-471.

scripts/generate_openapi_client.sh
kelvah: Uhm do you think we should drop the suffix entirely?
I suggested to remove "API" from tags just to align to other services. But I think we could eventually drop the suffix entirely if it's possible. Wdyt?

kelvah (commented):
Hi @jomarko, I have mixed feelings about this PR. I mean it's good to regenerate the new SDK after the spec has changed but I don't think we solved all the issues.

Like you said, the model has improved but still has some weird interfaces.

We now have `operationId`s in the spec but their values are redundant, leading to something like `BridgesAPISdk.bridgesAPIGetBridge()` to retrieve bridges.

So how should we proceed? We merge this one and keep the ticket? We create a new one?

scripts/generate_openapi_client.sh
jomarko: @kelvah I see your point. In my updates of sandbox openapi annotation this was included. It seems not in Jacopo's version of annotations. So I asked him https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/436/files#r832319124

jomarko:
@kelvah well I am fine with both approaches. If MGDOBR-420 is not blocking you form other work, I I am fine to keep it open, and addressing your comments.

vpellegrino (approved):
LGTM, thank you @jomarko

PR #9, opened 2022-03-22
[MGDOBR-413] Configure keycloak authentication

kelvah:
https://issues.redhat.com/browse/MGDOBR-413

This PR adds keycloak authentication to the standalone UI app to allow local API integration.

The keyclock config will change after sso will be added ([see 466](https://issues.redhat.com/browse/MGDOBR-466)). 

After that it will be possible to test the authentication with the SE as federated module.

Steps to test it:

- `npm install`
- `npm run start`

### DIFF
--- a/src/Keycloak.tsx
+++ b/src/Keycloak.tsx
@@ -1,0 +1,108 @@
+import Keycloak from 'keycloak-js';
+import React from 'react';
+
+import { Auth, AuthContext } from '@rhoas/app-services-ui-shared';
+
+export let keycloak: Keycloak.KeycloakInstance | undefined;
+
+/**
+ * Get keycloak instance
+ *
+ * @return an initiated keycloak instance or `undefined`
+ * if keycloak isn't configured
+ *
+ */
+export const setKeycloakInstance = async () => {
+  if (!keycloak) await init();
+};
+
+/**
+ * Initiate keycloak instance.
+ *
+ * Set keycloak to undefined if
+ * keycloak isn't configured
+ *
+ */
+export const init = async () => {
+  try {
+    /* Keycloak configuration using OB authentication.
+     * Thiese params will change after sso is implemented.
+     * See https://issues.redhat.com/browse/MGDOBR-466
+     */
+    keycloak = new (Keycloak as any)({
+      realm: 'event-bridge-fm',
+      url: 'https://keycloak-event-bridge-prod.apps.openbridge-dev.fdvn.p1.openshiftapps.com/auth/',
+      clientId: 'event-bridge',
+    });
+    if (keycloak) {
+      await keycloak.init({
+        onLoad: 'login-required',
+      });
+    }
+  } catch (e) {
+    keycloak = undefined;
+    console.warn(
+      'Auth: Unable to initialize keycloak. Client side will not be configured to use authentication',
+      e
+    );
+  }
+};
+
+/**
+ * Use keycloak update token function to retrieve
+ * keycloak token
+ *
+ * @return keycloak token or empty string if keycloak
+ * isn't configured
+ *
+ */
+export const getKeyCloakToken = async (): Promise<string> => {
+  await keycloak?.updateToken(50);
+  if (keycloak?.token) return keycloak.token;
+  console.error('No keycloak token available');
+  return 'foo';
+};
+
+/**
+ * Use keycloak update token function to retrieve
+ * keycloak token
+ *
+ * @return keycloak token or empty string if keycloak
+ * isn't configured
+ *
+ */
+export const getParsedKeyCloakToken =
+  async (): Promise<Keycloak.KeycloakTokenParsed> => {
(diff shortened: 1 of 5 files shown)


src/Keycloak.tsx
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Use keycloak update token function to retrieve
+ * keycloak token
+ *
+ * @return keycloak token or empty string if keycloak
+ * isn't configured
+ *
+ */
+export const getKeyCloakToken = async (): Promise<string> => {
+  await keycloak?.updateToken(50);
```
vpellegrino, line 60: what is `50`?

src/Keycloak.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * Use keycloak update token function to retrieve
+ * keycloak token
+ *
+ * @return keycloak token or empty string if keycloak
+ * isn't configured
+ *
+ */
+export const getParsedKeyCloakToken =
+  async (): Promise<Keycloak.KeycloakTokenParsed> => {
+    await keycloak?.updateToken(50);
```
vpellegrino, line 76: same question as above

src/Keycloak.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * keycloak isn't configured
+ *
+ */
+export const init = async () => {
+  try {
+    /* Keycloak configuration using OB authentication.
+     * Thiese params will change after sso is implemented.
+     * See https://issues.redhat.com/browse/MGDOBR-466
+     */
+    keycloak = new (Keycloak as any)({
```
vpellegrino, line 32: I would avoid `any` in favor of `unknown` or better specify an interface for it.

src/Keycloak.tsx
```diff
@@ -0,0 +1,108 @@
...
+ */
+export const logout = async () => {
+  if (keycloak) {
+    await keycloak.logout();
+  }
+};
+
+export const KeycloakAuthProvider: React.FunctionComponent = (props) => {
+  const getUsername = () => {
+    return getParsedKeyCloakToken().then((token: any) => token['username']);
```
vpellegrino, line 94: I think we can avoid using `any`, as I wrote above

vpellegrino (changes requested):
@kelvah thanks for this PR.
I tried it locally and it works fine.
Just a few minor comments to accomplish, before approving.

src/Keycloak.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * Initiate keycloak instance.
+ *
+ * Set keycloak to undefined if
+ * keycloak isn't configured
+ *
+ */
+export const init = async () => {
+  try {
+    /* Keycloak configuration using OB authentication.
+     * Thiese params will change after sso is implemented.
```
manstis, line 29: Typo `these`

manstis (commented):
A really, really, really minor comment!

jomarko:
Just FYI, I will review once comments above are addressed to not mix comments too much.

src/Keycloak.tsx
kelvah: It's the token minimum validity in seconds. If the token is going to expire within 50s it will be refreshed.

src/Keycloak.tsx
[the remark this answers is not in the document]
kelvah: Done.
Good point btw. We could consider adding the no-any rule to es-lint.

src/Keycloak.tsx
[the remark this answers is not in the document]
kelvah: Done.
The username retrieval looks a bit shady right now.
We don't need to get it ATM, let's wait for the sso integration to check if that's the right way to retrieve it.

src/Keycloak.tsx
kelvah: Whoops. Thanks for spotting it! Fixed now.

src/Keycloak.tsx
vpellegrino: I expect to have a constant with a meaningful name for it or at least a comment which explains why 50 it's the correct value and not 60, 70, 80, 90, 100...

vpellegrino (approved):
Great job! :rocket: 
Thank you @kelvah

src/Keycloak.tsx
kelvah: Fair point :+1:  

I moved the value to a constant and added a description there.

There's no "correct" value. The only important thing is it should be lower than the token lifespan. For our keycloak instance the lifespan is very high, 10 hours. It will be 15 minutes with sso.

To clarify, this keycloak setup is only used for local development of the standalone SE app.
When the SE module is imported within app services, it will retrieve tokens from there using the auth hook.

jomarko (approved):
Thank you, it seems to work both on chrome and firefox.

manstis (approved):
LGTM :+1:

PR #10, opened 2022-03-22
[MGDOBR-417]: Create a generic Table component

vpellegrino:
https://issues.redhat.com/browse/MGDOBR-417

With this PR, the `InstancesListPage` has been updated, to show how the `Table` component will be integrated. However, consider that the page itself will be managed later: https://issues.redhat.com/browse/MGDOBR-450

image


src/app/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+  test('should render configured columns', () => {
+    const { container } = render(<Table columns={columns} rows={[]} />);
+
+    const table = container.querySelector('table');
+    expect(table).not.toBeNull();
+    const thead = table?.querySelector('thead');
+    expect(thead).not.toBeNull();
+    const headers = thead?.querySelectorAll('th');
+    expect(headers).toHaveLength(columns.length);
+    expect(headers?.[0]).toHaveTextContent(columns[0].label);
```
manstis, line 19: Arguably you should also include `columns[1]` too.

manstis (approved):
LGTM, just one minor comment.

vpellegrino:
> Hi @vpellegrino, quick question before a proper review.
> 
> I see you use the "legacy" PF table. The [docs](https://www.patternfly.org/2021.16/components/table) say it will still be maintained but new features will only be added to the "default" `TableComposable`.
> 
> Any particular reason behind your choice? Do you think `Table` is better?

Hi @kelvah.

I've used `TableComposable` in the past and it led to really complex JSX and complex configuration. It was suggested to me when I wanted to integrate react-table library inside that project. 
Since "Both implementations are supported and fully maintained" I decided to use the simple/basic `Table` which seems more aligned to our mockups/use-cases.
Let me add that the fact that we are wrapping PF Table inside a specific component, will make easy any change we are committed to do in the future, in one shot. 
However, if you think we should switch to `TableComposable`, no issue from my side. Just let me know your opinion.

kelvah:
@vpellegrino Thanks for clarifying. 

I think developer experience should not overcome user experience. If choosing a "legacy" tool for ease-of-development could prevent us from bringing new features to the user, then I consider it a bad idea.

This is a general rule, of course. In our context we are talking about a... table. It's not the most complex UI element.

We also do not have particular requirements ATM. Just a table doing table things.

So no big deal right now If you want to stay with `Table`. We can move over to other things.

vpellegrino:
> @vpellegrino Thanks for clarifying.
> 
> I think developer experience should not overcome user experience. If choosing a "legacy" tool for ease-of-development could prevent us from bringing new features to the user, then I consider it a bad idea.
> 
> This is a general rule, of course. In our context we are talking about a... table. It's not the most complex UI element.
> 
> We also do not have particular requirements ATM. Just a table doing table things.
> 
> So no big deal right now If you want to stay with `Table`. We can move over to other things.

@kelvah looking at your comment, I am a bit worried that you could think that adopting `Table` can _potentially_ lead to a worse user experience. 
I need to clarify that I am always willing to provide the best user experience possible, as the first factor I consider while developing is, indeed, the user.

So, I am going to change `Table` to `TableComposable` as you require.

src/app/components/Table/Table.tsx
```diff
@@ -0,0 +1,98 @@
...
+        </Tr>
+      </Thead>
+      <Tbody>
+        {transformRows(rows, columns).map((row, rowIndex) => (
+          <Tr key={row?.originalData?.id ?? rowIndex}>
+            {row?.cells?.map((cell, cellIndex) => (
+              <Td key={cellIndex}>{cell}</Td>
+            ))}
+            {actionResolver && (
+              <Td className="pf-c-table__action">
```
jomarko, line 88: What PF version do we use? Can't we just set `isActionCell` ?

https://www.patternfly.org/v4/components/table#td

jomarko (commented):
Thank you for a PR. One question about a code.

src/app/components/Table/Table.tsx
vpellegrino: I already tried. Not able to set it:
image
We are using:

```
    "@patternfly/patternfly": "4.164.2",
    "@patternfly/react-core": "4.181.1",
    "@patternfly/react-icons": "4.32.1",
    "@patternfly/react-styles": "4.31.1",
    "@patternfly/react-table": "4.50.1",
```
It seems that `isActionCell` will set only the `CSS` class, looking at the documentation.

Anyway, even if you set `isActionCell`, you need to manually wrap `<ActionsColumn` inside the `td`, while using `TableComposable`.
image

jomarko: Ok, to me it seems it was added by this commit into patternfly

https://github.com/patternfly/patternfly-react/commit/eaaedc28cf54cca1a47aaa5c4f5f76a11d32864d

So we are using PF where the feature is not available?

vpellegrino: Correct @jomarko. It seems we are using a version where this attribute is not defined.

jomarko: is it worth to comment in code or report a jira, so we can "fix/use the feature" in future, when we upgrade `patternfly/react-table` library?

vpellegrino: In my opinion no, because `isActionCell` adds only one CSS class (`pf-c-table__action`) to `td` element:
image

jomarko (approved):

src/app/components/Table/Table.tsx
kelvah: my 2 cents, I agree with @jomarko. If we know there's a `isActionCell` prop coming to do the job, we should track a ticket to swap to it.

I don't know if it's the right thing to do but I see there's another option, to add actions with the `actions` prop of `Td`.
It will add the action css class but it's also adding other inline styles, so aesthetically it's not the same thing right now. It may get the correct right alignment with more recent versions of PF though.

By the way the PF version we currently use is `2021.16`

kelvah (commented):
Thank you @vpellegrino, I've left a couple of comments.

package-lock.json
```diff
@@ -1426,9 +1426,9 @@
       }
     },
     "node_modules/@redhat-cloud-services/frontend-components-config": {
-      "version": "4.6.2",
-      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/frontend-components-config/-/frontend-components-config-4.6.2.tgz",
-      "integrity": "sha512-Jlhvwor7DpiCE+JyFTl9ksFQ5/lpXw5Lnp0ZTim38KccjAxlL5Do5NBlz6o8kGMikeECUejVBbTOFpzrBVWzHA==",
+      "version": "4.6.4",
```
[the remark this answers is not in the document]
vpellegrino: It was suggested by `npm audit fix`. Anyway, we still have several security warnings that deserve a separate investigation. I will revert it

src/app/components/Table/Table.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /** Element to be appended after the `tbody` node */
+  children?: React.ReactNode;
+  /** Columns to display */
+  columns: TableColumn[];
+  /** list of additive css classes */
+  cssClasses?: string | string[];
+  /** Collection of cells to render */
+  rows: IRow[];
+  /** Style variant for the table */
+  variant?: 'compact' | undefined;
```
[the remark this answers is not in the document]
vpellegrino: I was wondering why the PF team is exposing such property then. Will remove it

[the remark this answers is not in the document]
vpellegrino: @kelvah, just tried it out.
Compact:
image
w/o compact:
image

The mockups seem to be an intermediate version of these two.
I personal prefer the Compact version. Anyway it makes sense to have such property configurable. WDYT?

src/app/components/Table/Table.tsx
vpellegrino: Traced a defect: https://issues.redhat.com/browse/MGDOBR-489

vpellegrino:
@kelvah accomplished all your feedbacks.
I am just waiting for your answer to https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/10#discussion_r834257488

src/app/components/Table/Table.tsx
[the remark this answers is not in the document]
kelvah: > The mockups seem to be an intermediate version of these two. I personal prefer the Compact version. Anyway it makes sense to have such property configurable. WDYT?

Let's ask Susan which version to use for confirmation.

Yes, it makes sens to keep it configurable. I wasn't suggesting to remove the prop by the way :sweat_smile: 
It just needs to be fixed. There's no way right now to say to our Table component to use the default variant, am I right?

src/app/components/Table/Table.tsx
kelvah: :+1:

src/app/components/Table/Table.tsx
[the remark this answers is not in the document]
vpellegrino: @kelvah now I got the point of your comment. Thank you for catching it.
Fixed: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/10/commits/f1dc1174402e9eff693f4d94d0b3db6789518578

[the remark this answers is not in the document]
vpellegrino: About compact vs no-compact, I will ask Susan for confirmation and I will accomplish her feedback in https://issues.redhat.com/browse/MGDOBR-450

src/app/components/Table/Table.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /** Element to be appended after the `tbody` node */
+  children?: React.ReactNode;
+  /** Columns to display */
+  columns: TableColumn[];
+  /** list of additive css classes */
+  cssClasses?: string | string[];
+  /** Collection of cells to render */
+  rows: IRow[];
+  /** Style variant for the table */
+  variant?: "compact" | undefined;
```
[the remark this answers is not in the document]
vpellegrino: You're right! Thank you!

PR #11, opened 2022-03-22
MGDOBR-444: Introduce Smart Events UI CI quality checks

jomarko:
The github action needs to be checked on my fork: https://github.com/jomarko/sandbox-ui/actions


jomarko:
@kelvah @manstis @vpellegrino as this PR demonstrates, our current prettier config is set to have 'strings' vs "strings". To me are double quotes more natural (due to my java background), just wanted to know your opinion.

manstis:
> @kelvah @manstis @vpellegrino as this PR demonstrates, our current prettier config is set to have 'strings' vs "strings". To me are double quotes more natural (due to my java background), just wanted to know your opinion.

I have no strong opinion either way. I "grew up" with the format `"here is a 'string' inside a 'string'"` from my JS days too. However `'here is a "string" inside a "string"'` works equally well and seems to be the standard inside kiegroup repositories for TypeScript code (interestingly the prettier default seems to be "dynamic": https://prettier.io/docs/en/rationale.html#strings. Perhaps we force single quotes to avoid a mix-and-match approach that could result from dynamic formatting).

vpellegrino:
> @kelvah @manstis @vpellegrino as this PR demonstrates, our current prettier config is set to have 'strings' vs "strings". To me are double quotes more natural (due to my java background), just wanted to know your opinion.

I'd prefer to avoid mixed approaches, for sure.
My personal preference is about having double quotes: "string"

jomarko:
I will do double quotes our default setting then

kelvah:
> I will do double quotes our default setting then

I'm late to the party but I don't have a strong opinion. it's OK for me to change it.

Just to clarify, the prettier quotes "conversion" will stay dynamic regardless of `singleQuote`. It will still convert strings to single quotes to remove escapes when they are present.

jomarko:
I believe the CI setup is worth to review now, please notice, I would like to review eslint config in separate ticket https://issues.redhat.com/browse/MGDOBR-478

.github/workflows/quality-checks.yml
```diff
@@ -0,0 +1,15 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14"
+          cache: "npm"
+      - run: npm install
+      - run: npm run format:check
+      - run: npm run lint
+      - run: npm run test
```
vpellegrino, line 15: is one of these commands already covering type checking?

vpellegrino (commented):
PR looks fine @jomarko. Just one single question (double-check) before approving

.github/workflows/quality-checks.yml
jomarko: I tend to say no, beacuse of https://issues.redhat.com/browse/MGDOBR-478, may I ask you to add such a requirement into linked jira?

vpellegrino: Ok, I updated the jira. Thank you

vpellegrino (approved):
Approving, thank you @jomarko

kelvah (approved):
LGTM :+1:

PR #14, opened 2022-03-31
MGDOBR-418: "Create Smart Events instance" template 

kelvah:
https://issues.redhat.com/browse/MGDOBR-418

It follows https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/13 to use the test setup available there.

If you are reviewing this PR before #13 is merged, please filter only the last commit (_"MGDOBR-418: add CreateInstance component"_)

Steps to open it:
- `npm run start`
- login
- click on "Create Smart Events instance"

### DIFF
--- a/src/app/Instance/CreateInstance/CreateInstance.tsx
+++ b/src/app/Instance/CreateInstance/CreateInstance.tsx
@@ -1,0 +1,124 @@
+import React, { FormEvent, useCallback, useEffect, useState } from "react";
+import {
+  Alert,
+  AlertGroup,
+  Button,
+  Form,
+  FormGroup,
+  Modal,
+  TextInput,
+} from "@patternfly/react-core";
+import { useTranslation } from "react-i18next";
+
+export interface CreateInstanceProps {
+  /**
+   * Flag to indicate the creation request is in progress
+   */
+  isLoading: boolean;
+  /**
+   * Flag to show/hide the modal
+   */
+  isModalOpen: boolean;
+  /**
+   * Callback to close the modal
+   */
+  onClose: () => void;
+  /**
+   * Callback to create the instance
+   */
+  onCreate: (name: string) => void;
+}
+
+const CreateInstance = (props: CreateInstanceProps) => {
+  const { isLoading, isModalOpen, onClose, onCreate } = props;
+  const [name, setName] = useState("");
+  // @TODO Basic temporary error management. To be integrated with APIs errors.
+  const [error, setError] = useState<string | null>(null);
+  const { t } = useTranslation("openbridgeTempDictionary");
+
+  const FORM_ID = "create-instance-form";
+
+  const onSubmit = useCallback(
+    (event: FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+      if (name === "") {
+        setError("name-required");
+      } else {
+        setError(null);
+        onCreate(name);
+      }
+    },
+    [name, onCreate]
+  );
+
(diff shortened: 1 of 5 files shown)


jomarko (commented):
Thank you for a pr @kelvah . I have one question. why we need both
- 'Create a Smart Events instance'
- 'Create Smart Events instance'

Screenshot from 2022-04-01 08-37-34


To me it seems 'Smart Events' are over used / over mentioned. I think you touch that topic on meeting already. From my point of view user has context when he begins some interaction.

So for example if the button on landing screen is "Create Smart Events instance", the dialog button would be enough to be 'Create', as user already knows hat he is creating:
- the button on landing screen advised
- also the opened dialog title advises this

To me it seems something as to be revisited. But it is fail to say, I am not UX expert. What is the opinion of others?

kelvah:
> Thank you for a pr @kelvah . I have one question. why we need both
> 
> * 'Create a Smart Events instance'
> * 'Create Smart Events instance'
> 
> [...]
> 
> To me it seems something as to be revisited. But it is fail to say, I am not UX expert. What is the opinion of others?

I agree with you on this. I raised the topic again with UX. You can find some comments on this in the slide deck mentioned in the jira ticket description (slide 5).

To summarize, we are aligning with what's happening on other services. The shared pattern is to use the naming `Create [product name] instance`. There could be also scenarios where you can create an instance of a service from the UI of another one. In this case you are in another context, so having just "instance" or "create" could be not enough. (yes, we could have a solution to that, configurable labels provided by props :slightly_smiling_face: )

So, for now, we are keeping the _redundant_ names. it's always true that this is the very first version of the UI. We will iterate on it, collecting feedback from wider audiences as soon as we have something to show. So nothing is set in stone.

jomarko (approved):
thank you @kelvah

vpellegrino (approved):
Review the commit https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/14/commits/edd372f2c51b7c50ad9cc5634a13beee36d08c71
good for me

manstis:
@kelvah I was going to review when you've rebased.

Could you please let me know when you've done it.

kelvah:
> @kelvah I was going to review when you've rebased.
> 
> Could you please let me know when you've done it.

@manstis done

src/app/Instance/CreateInstance/CreateInstance.tsx
```diff
@@ -0,0 +1,124 @@
...
+        setError(null);
+        onCreate(name);
+      }
+    },
+    [name, onCreate]
+  );
+
+  const handleNameChange = useCallback((value: string) => {
+    setName(value);
+    setError(value === "" ? "name-required" : null);
```
manstis, line 56: Is there some App Services related explanation for this magic string and subsequent use on L#59?

manstis (approved):
A question about why you do something, but it doesn't affect my approval.

src/app/Instance/CreateInstance/CreateInstance.tsx
kelvah: No. this is only a magic string.  I added a comment on L#35 about it. It's only a temporary basic validation. I think the real one will be plugged to the APIs and will refer to error codes coming from there. So we'll reuse the "name missing" error code when preventing the form submission when the name input is empty.
Also maybe the validation will be handled outside the `CreateInstance` component and passed down as a prop? We have to discuss it properly.

PR #15, opened 2022-04-01
MGDOBR-478: Improve eslint checks

jomarko:


jomarko:
Oh, force push, I lost the Valentinos commit, I need to fix it.

src/App.tsx
```diff
@@ -19,9 +19,11 @@ const App = () => {
   useEffect(() => {
     const init = async () => {
       await setKeycloakInstance();
-      setInitialized(true);
     };
-    init();
+    init().then(
+      () => setInitialized(true),
+      () => setInitialized(false)
+    );
```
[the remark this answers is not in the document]
vpellegrino: yes, you can just rollback your changes and change line 24 with:
```
    void init();
```

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -8,23 +8,27 @@ import {
...
       label: "Name",
-      formatter: (value: IRowData, row?: IRow) => (
-        <Link
-          data-testid="tableInstances-linkInstance"
-          to={`instance/${row?.id}`}
-        >
-          {value}
-        </Link>
-      ),
+      formatter: (value: IRowData, row?) => {
```
manstis, line 21: It seems weird that we're mandating "typing" e.g. [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/15/files#diff-0e766af1b39216dbeff5e7d8c1965bf8ad708f1c98f9ff42edb1ee55197977c2R17) but on this line we're actually removing the typing and casting on the next line.

src/app/components/Table/Table.tsx
```diff
@@ -79,8 +85,8 @@ export const Table: FunctionComponent<TableProps> = ({
         </Tr>
       </Thead>
       <Tbody>
-        {transformRows(rows, columns).map((row, rowIndex) => (
-          <Tr key={row?.originalData?.id ?? rowIndex}>
+        {transformRows(rows, columns).map((row: TableRow, rowIndex) => (
```
manstis, line 88: Why do we not need "typing" on `rowIndex`?

src/app/Instance/InstancesListPage/InstancesListPage.tsx
vpellegrino: @manstis I am the author of this change.
The type of `row` is inferred to be `IRow` (it is not changed).
On line 22 I casted to `Instance` to give it more context related to this page (and in order to access to `id` property).
In my opinion it was not needed to have the type explicit in the function argument, since it is casted in line 22.
But feel free to restore it, if you prefer.

src/App.tsx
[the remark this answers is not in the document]
jomarko: done, could you please explain this construct a little bit?

src/app/components/Table/Table.tsx
vpellegrino: Because from `.map` function definition you already know that the second argument is always a number, so the compiler is able to infer it and it is not necessary to declare it.
However, if you prefer I am ok for adding it explicit.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
manstis: @vpellegrino OK. I thought we were mandating use of type information on this PR with linting rules.

If this is not the case then please ignore my comment.

manstis (approved):
From what I understand, LGTM.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
jomarko: I added `row?: IRow` in a separate commit

src/App.tsx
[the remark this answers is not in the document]
vpellegrino: From what I understand, the linting wants to make sure that every `Promise` is awaited or, if you are not interested in observing the Promise result, declare it as `void`.
It is a rule to make developers more conscious about async code.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
vpellegrino: @manstis fair point. I was trying to give more context. I am always willing to learn new things and make the code more easy to read.

.eslintrc.js
```diff
@@ -1,11 +1,80 @@
...
+    "jest-dom",
+    "react",
+    "react-hooks",
+    "testing-library",
+    //"xstate", cos-ui https://issues.redhat.com/browse/MGDOBR-510
+  ],
+
+  // kafka-ui
+  rules: {
+    "@typescript-eslint/no-unused-vars": "off",
```
kelvah, line 34: Weeeeell no unused vars. Always a pleasure to discuss it. :slightly_smiling_face: 

I prefer not having errors on unused vars (and imports) while developing because it can be a pain while you are trying solutions to always comment stuff before saving and rebuilding. At the same time it's nice to not leave unused stuff when committing code.

So turning this off on es lint would probably imply also turning off the `noUnusedLocals` from ts config. Otherwise we have a mix, compilation errors but no linting errors.

A compromise is to have `noUnusedLocals` turned off and keep `no-unused-vars` but only as a warning. So you are free to try things during development and then you have the chance to remove unused stuff to make linting warnings go away before submitting a PR. So with this approach we are trusting devs to remember to get rid of the warnings.

The alternative is to keep both eslint and tsconfig screaming on unused vars.

kelvah (approved):
Thanks @jomarko. Left some comments.

.eslintrc.js
```diff
@@ -1,11 +1,80 @@
...
+    "@typescript-eslint/explicit-function-return-type": "off", // https://issues.redhat.com/browse/MGDOBR-530
```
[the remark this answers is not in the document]
jomarko: sure, lets discuss it separately from this PR then, in jira MGDOBR-510, possibly we will just close it and remove this ticket reference form codebase

.eslintrc.js
```diff
@@ -1,11 +1,80 @@
...
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: ["./tsconfig.json"],
+  },
   extends: [
+    "plugin:jest/recommended",
     "plugin:jest-dom/recommended",
+    "plugin:react/recommended",
     "plugin:react-hooks/recommended",
+    "plugin:react/recommended",
```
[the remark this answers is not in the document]
jomarko: good catch, removed in a separate commit

.eslintrc.js
jomarko: I am happy to turn it on.

jomarko:
Merging this, to have checks for upcoming PRs. We can still adjust in future.

PR #16, opened 2022-04-04
MGDOBR-450: Create bridge instances overview page

vpellegrino:
Created the **static version** of bridge instances page, aligned to mockup, except for:
- Missing sort and filtering features, since it is not clear when the related endpoints will be ready, and which columns will be sorted and filtered (and how).
- Status column is not displayed as expected, because it is not currently possible to reuse [`KafkaInstanceStatus`](https://redhat-developer.github.io/app-services-ui-components/?path=/docs/kafka-kafka-instance-status-kafka-instance-status--accepted) component (1) since it seems that this component is not generic as expected and we cannot customize contained labels.

(1):
By declaring:
```
    {
      accessor: "status",
      label: t("instance.status"),
      formatter: (value: IRowData, row?: IRow) => {
        return (
          <KafkaInstanceStatus
            status={value as unknown as KafkaStatus}
            createdAt={new Date(row?.["submitted_at"])}
            onClickConnectionTabLink={() =>
              // @TODO missing action when clicking on the connection tab link
              {}
            }
            onClickSupportLink={() =>
              // @TODO missing action when clicking on the support link
              {}
            }
          />
        );
      },
    },
```
we have:
image

### DIFF
--- a/locales/en/openbridge.json
+++ b/locales/en/openbridge.json
@@ -3,14 +3,17 @@
     "ago": "ago",
     "cancel": "Cancel",
     "create": "Create",
-    "details": "Details",
     "delete": "Delete",
+    "description": "Description",
+    "details": "Details",
     "endpoint": "Endpoint",
     "kafkaTopic": "Kafka Topic",
     "key": "Key",
     "name": "Name",
     "required": "Required",
     "selectType": "Select type",
+    "status": "Status",
+    "submittedAt": "Time created",
     "submittingRequest": "Submitting request",
     "type": "Type",
     "value": "Value",
@@ -21,13 +24,9 @@
   "instance": {
     "createASEInstance": "Create a Smart Events instance",
     "createSEInstance": "Create Smart Events instance",
-    "description": "Description",
     "instanceId": "Instance ID",
     "instanceWillBeAvailableShortly": "Your Smart Events instance will be ready for use shortly after creation.",
-    "name": "Name",
-    "smartEventInstances": "Smart Event Instances",
-    "status": "Status",
-    "submittedAt": "Time created"
+    "smartEventInstances": "Smart Event Instances"
   },
   "processor": {
     "action": "Action",
--- a/locales/en/openbridge.json
+++ b/locales/en/openbridge.json
@@ -5,7 +5,10 @@
     "endpoint": "Endpoint",
     "kafkaTopic": "Kafka Topic",
     "key": "Key",
+    "name": "Name",
+    "required": "Required",
     "selectType": "Select type",
+    "submittingRequest": "Submitting request",
     "type": "Type",
     "value": "Value",
     "webhook": "Webhook"
@@ -13,8 +16,11 @@
   "demoOverview": "Demo Overview",
   "instancesListTable": "Instances list table",
   "instance": {
+    "createASEInstance": "Create a Smart Events instance",
+    "createSEInstance": "Create Smart Events instance",
     "smartEventInstances": "Smart Event Instances",
-    "instanceId": "Instance ID"
+    "instanceId": "Instance ID",
+    "instanceWillBeAvailableShortly": "Your Smart Events instance will be ready for use shortly after creation."
   },
   "processor": {
     "action": "Action",
--- a/locales/en/openbridge.json
+++ b/locales/en/openbridge.json
@@ -1,7 +1,10 @@
 {
   "common": {
+    "ago": "ago",
     "cancel": "Cancel",
     "create": "Create",
+    "details": "Details",
+    "delete": "Delete",
     "endpoint": "Endpoint",
     "kafkaTopic": "Kafka Topic",
     "key": "Key",
@@ -18,9 +21,13 @@
   "instance": {
     "createASEInstance": "Create a Smart Events instance",
     "createSEInstance": "Create Smart Events instance",
-    "smartEventInstances": "Smart Event Instances",
+    "description": "Description",
     "instanceId": "Instance ID",
-    "instanceWillBeAvailableShortly": "Your Smart Events instance will be ready for use shortly after creation."
+    "instanceWillBeAvailableShortly": "Your Smart Events instance will be ready for use shortly after creation.",
+    "name": "Name",
+    "smartEventInstances": "Smart Event Instances",
+    "status": "Status",
+    "submittedAt": "Time created"
   },
   "processor": {
     "action": "Action",
--- a/src/app/Instance/InstancesList/InstancesList.tsx
+++ b/src/app/Instance/InstancesList/InstancesList.tsx
@@ -61,7 +61,7 @@ export const InstancesList: FunctionComponent<InstancesListProps> = ({
 
   return (
     <Card>
-      <Toolbar>
+      <Toolbar ouiaId="instances-toolbar">
         <ToolbarContent>
           <ToolbarItem alignment={{ default: "alignLeft" }}>
             <Button onClick={() => setShowCreateInstance(true)}>
--- a/src/app/Instance/InstancesList/InstancesList.tsx
+++ b/src/app/Instance/InstancesList/InstancesList.tsx
@@ -1,0 +1,106 @@
+import React, { FunctionComponent, useState } from "react";
+import {
+  Button,
+  Card,
+  Toolbar,
+  ToolbarContent,
+  ToolbarItem,
+} from "@patternfly/react-core";
+import CreateInstance from "@app/Instance/CreateInstance/CreateInstance";
+import { Table, TableColumn } from "@app/components/Table";
+import { Instance } from "../../../types/Instance";
+import { Pagination } from "@app/components/Pagination/Pagination";
+import { useTranslation } from "react-i18next";
+
+interface InstancesListProps {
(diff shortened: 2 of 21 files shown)


manstis (approved):
A couple of minor comments about inlining CSS, but IDK if we have any alternative.

manstis:
@vpellegrino what is our plan for re-use of `KafkaInstanceStatus`?

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -128,25 +173,37 @@ const InstancesListPage = () => {
...
+              {getPagination(true)}
+            </section>
+          }
           actionResolver={actionResolver}
           ariaLabel={t("openbridgeTempDictionary:instancesListTable")}
           columns={columnNames}
           cssClasses="tableInstances"
           rows={instances}
         />
+        {getPagination(false, {
```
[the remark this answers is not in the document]
vpellegrino: I'm not sure which style support we want to adopt (CSS, SCSS, ...) && this page is going to change very frequently, so I took this road. I want to listen to your opinion guys, what do we want to use? plain CSS?

vpellegrino:
> @vpellegrino what is our plan for re-use of `KafkaInstanceStatus`?

We can decide together.
IMO, We have several options:
- Write our own `InstanceStatus`
- At a certain point `KafkaInstanceStatus` will be made more general
- Open a PR against the repo containing `KafkaInstanceStatus`

Do you see other options? Anyway, I think we can postpone this a bit, also to make sure that the `smart-events` status will match perfectly the `Kafka` ones.
WDYT @kelvah?

src/app/Instance/InstancesListPage/InstancesListPage.tsx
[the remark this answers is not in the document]
manstis: Fair point.. "we" (@kelvah and myself) gravitate to including class names not (S)CSS itself.

manstis:
@vpellegrino Any approach is fine.. I think the important thing is that it's not forgotten about.

A JIRA for us for a later date, a JIRA/issue for `app-services-ui-components` etc.. as long as we don't forget.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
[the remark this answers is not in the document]
kelvah: @vpellegrino  +1 to moving CSS to separate files. Just plain CSS, no sass. Also I've started using [BEM](http://getbem.com/introduction/) as naming convention in other PR's, so I would say let's keep the same approach.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -128,25 +173,37 @@ const InstancesListPage = () => {
...
+                {t("instance.createSEInstance")}
+              </Button>
+              <CreateInstance
+                isLoading={false}
+                isModalOpen={showCreateInstance}
+                onClose={() => setShowCreateInstance(false)}
+                onCreate={() => setShowCreateInstance(false)}
+              />
+              {getPagination(true)}
+            </section>
```
kelvah, line 195: I see you are using the caption for the controls. There's nothing technically wrong with it but I think a PF Toolbar is a better fit. [Here's an example](https://www.patternfly.org/v4/components/table/react-demos#bulk-select).
It will also take care of the correct spacing and responsiveness.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -26,99 +30,140 @@ const InstancesListPage = () => {
...
+        onClick: () =>
+          // @TODO missing action to perform when clicking on delete action
+          {},
       },
     ];
   };
 
   const [showCreateInstance, setShowCreateInstance] = useState(false);
 
+  const getPagination = (isCompact: boolean, customStyle?: CSSProperties) => (
```
kelvah, line 153: Minor observation given this will change. I would probably pass the `itemCount` too as param.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { CSSProperties, useState } from "react";
```
kelvah, line 1: I think we should split the page components from the UI components we plan to move.
Like `InstancesListPage` and `InstancesList`. The first one will stay here and take care of logic and data fetch. The second one will be the UI component moved over (table + paginations).
Does it make sense?

kelvah (commented):
Thank you @vpellegrino. I left you some comments. 

I know we don't have mockups for the loading state of this page. Should we create a different jira to manage it?

vpellegrino:
> * Status column is not displayed as expected, because it is not currently possible to reuse [`KafkaInstanceStatus`](https://redhat-developer.github.io/app-services-ui-components/?path=/docs/kafka-kafka-instance-status-kafka-instance-status--accepted) component (1) since it seems that this component is not generic as expected and we cannot customize contained labels.

@manstis / @kelvah traced https://issues.redhat.com/browse/MGDOBR-523

locales/en/openbridge.json
```diff
@@ -18,9 +21,13 @@
...
     "createSEInstance": "Create Smart Events instance",
-    "smartEventInstances": "Smart Event Instances",
+    "description": "Description",
     "instanceId": "Instance ID",
-    "instanceWillBeAvailableShortly": "Your Smart Events instance will be ready for use shortly after creation."
+    "instanceWillBeAvailableShortly": "Your Smart Events instance will be ready for use shortly after creation.",
+    "name": "Name",
+    "smartEventInstances": "Smart Event Instances",
+    "status": "Status",
+    "submittedAt": "Time created"
```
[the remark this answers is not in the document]
vpellegrino: To be honest, my intention was to separate the domain of instances from other ones. In that way, changing `instance.name` to "Bridge Name" would not affect other labels. Anyway, I'm going to accomplish your proposal.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
vpellegrino: I don't see too much value in using `<Toolbar>` right now:
image
As you can see, I need to add padding on the left and the right. `<caption>` seems quite a more natural solution for this case.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
vpellegrino: I think that this topic deserves a separate meeting. Please ping me when you're available.

vpellegrino:
> Thank you @vpellegrino. I left you some comments.
> 
> I know we don't have mockups for the loading state of this page. Should we create a different jira to manage it?

If we don't have JIRA for loading status, we should trace them, yes (maybe during tomorrow's sync-up meeting).
Just my 2 cents: it makes sense to work on loading status and async operations only when we have fake/real API to call; otherwise it is very likely to re-implement them.

jomarko:
From the ongoing discussion, not sure this is ready for my review. Please let me know. It seems the code will change.

vpellegrino:
@kelvah / @jomarko I think I have accomplished to all existing comments. Feel free to let me know if you have further observations

src/app/Instance/InstancesList/InstancesList.tsx
```diff
@@ -0,0 +1,106 @@
...
+        ariaLabel={t("openbridgeTempDictionary:instancesListTable")}
+        columns={columnNames}
+        cssClasses="tableInstances"
+        rows={instances}
+      />
+      {getPagination(
+        instances.length,
+        false,
+        "instances-list-page__pagination--bottom"
+      )}
```
kelvah, line 103: Thanks @vpellegrino for all the changes you already did.
Another minor observation from me. I only noticed now that the bottom pagination lacks some right padding, it looks too close to the right border of its container.
I believe you can solve this quickly without manually apply padding if you add `variant="bottom"` to the [PF Pagination component](https://www.patternfly.org/v4/components/pagination/#pagination) in this case.

vpellegrino: @kelvah thanks for your feedback. Added the `variant` as well.

kelvah: :tada: 

You can even get rid of the `instances-list-page__pagination--bottom` class now. All the padding should be already covered.

kelvah (approved):
Thanks @vpellegrino. LGTM.

vpellegrino:
@jomarko any thoughts to share about this PR? I'm working on a strictly related task and I would like to avoid potential merge conflicts.

jomarko (commented):
@vpellegrino thank you for the PR.

I would like to incorporate this change https://github.com/vpellegrino/sandbox-ui/pull/1. It is related to QE team intention to have OUIA compliance for Smart Events UI in future - https://issues.redhat.com/browse/MGDOBR-533.

Manual check questions:
- pagination page size can not be changed, do we track this somehow?

jomarko (approved):
Thank you @vpellegrino

PR #19, opened 2022-04-12
MSW exploration

kelvah:
**This is only an exploratory work to evaluate MSW; proper setup will follow in a different PR.**

Steps to run the mocked version of the app

- `npm install`
- `npm run start:mocked`

Steps to run the mocked API as a separate express server

- `npm install`
- `npm run start:mock-server` (starting the mocked server)
- `npm run start` (in another terminal window)

Docs

- [msw](https://mswjs.io/)
- [msw data](https://github.com/mswjs/data)


src/App.tsx
```diff
@@ -24,6 +25,15 @@ const App = () => {
...
 
+  if (process.env.MOCKED_API) {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires,jest/no-mocks-import
+    const { worker } = require("./../__mocks__/msw/browser") as {
+      worker: SetupWorkerApi;
+    };
+
+    void worker.start();
+  }
+
```
vpellegrino, line 36: I think that you can remove this part from the `App` and put it in a separate nodescript to launch with the `start:mocked` command.

kelvah: I'm not following you :thinking:.

If we set up the worker like I'm doing here but in a node script it won't be reachable in the browser.

Or maybe you are suggesting to set it up like a "standalone" express server and launch it with a separate npm script? (Something similar to what we discussed yesterday.)

I'm not very fit today so I could have got your comment totally wrong :slightly_smiling_face:

vpellegrino: Let me check out your branch locally. I will check if what I have in mind it's possible and maybe I can open a PR against your fork

vpellegrino: Well, I double-checked and you're right with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/19#discussion_r850679470
The only alternative we have here is to leverage on a standalone server (for instance, `import { setupServer } from "msw/node"`). But, at the end of the day, I expect that `App` is going to be used only on the dev environment so there is nothing wrong with launching a service worker from it.

kelvah:
I also added an express server to run the api mocks as standalone. See updated PR description.

manstis:
@kelvah Does this need review from me now given #24?

kelvah:
> @kelvah Does this need review from me now given #24?

I don't think it's necessary to review this PR. I put it up just to showcase how it works. It's fine to review only #24.

kelvah:
Closing as the proper work has been done in other PRs.

PR #22, opened 2022-04-14
MGDOBR-452: Create instance drawer details

vpellegrino:
image


src/app/Instance/InstanceDetails/InstanceDetails.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <DescriptionListTerm>{t("common.submittedAt")}</DescriptionListTerm>
+            <DescriptionListDescription data-ouia-component-id="instance-details-submitted-date">
+              {formatDate(submitted_at)}
+            </DescriptionListDescription>
+          </DescriptionListGroup>
+          <DescriptionListGroup>
+            <DescriptionListTerm>{t("common.updatedAt")}</DescriptionListTerm>
+            <DescriptionListDescription data-ouia-component-id="instance-details-published-date">
+              {formatDate(published_at)}
+            </DescriptionListDescription>
```
kelvah, line 97: As you may have realized we don't have an "update date" for now. I would use a label reflecting our "Published at" value.

Unfortunately the drawer mockup is not up to date with the discussion we had inside comments.

src/app/Instance/InstancePage/InstancePage.tsx
```diff
@@ -126,102 +126,124 @@ const InstancePage = () => {
...
+        panelContent={
+          <InstanceDetails
+            onClosingDetails={() => setShowInstanceDrawer(false)}
+            id="3543edaa-1851-4ad7-96be-ebde7d20d717"
+            endpoint="https://ob-3543edaa-1851-4ad7-96be-ebde7d20d717.apps.openbridge-dev.fdvn.p1.openshiftapps.com/events"
+            name={instanceName}
+            owner="bebianco@redhat.com"
+            published_at="2022-04-12T12:06:22.881959+0000"
+            status="READY"
+            submitted_at="2022-04-12T12:04:43.044590+0000"
```
kelvah, line 141: We need almost all of the instance properties for this component. It's reasonable to pass the whole instance as a prop then instead of 6 different ones.

kelvah (commented):
Hi @vpellegrino, I've left only a couple of minor observations.

kelvah (approved):
Looking good :+1:

manstis:
Hi @vpellegrino could you please fix the conflict then I will review. Thank-you.

vpellegrino:
@manstis @jomarko can you please proceed with the review? Thank you

manstis (approved):
Two minor comments:-
- It's a shame the drawer is not wider as the endpoint URL property will always be truncated.
- Clicking on "Details" in the kebab menu on the Bridge Instances screen did not show the drawer.

vpellegrino:
> Two minor comments:-
> 
> * It's a shame the drawer is not wider as the endpoint URL property will always be truncated.
Then you would have a full-screen page instead of a drawer, and probably it will not be enough to show the full URL.
> * Clicking on "Details" in the kebab menu on the Bridge Instances screen did not show the drawer.
You have a point. I'm not sure we should accomplish it in this PR, anyway we could trace it (if not already traced), WDYT @kelvah ?

kelvah:
> 1. Then you would have a full-screen page instead of a drawer, and probably it will not be enough to show the full URL.
> 2. You have a point. I'm not sure we should accomplish it in this PR, anyway we could trace it (if not already traced), WDYT @kelvah ?

1. I've checked PF docs and it seems possible to increase the drawer width. See the [breakpoint example](https://www.patternfly.org/v4/components/drawer#breakpoint) where they set `<DrawerPanelContent widths={{ default: 'width_33' }}>`. I would give it a try.
2. I've missed it during my review. If it's not too much hassle I would do it here.

vpellegrino:
@kelvah @manstis 

image

This is how `width_66` looks like.
Please let me know if you like it and I will commit. Otherwise, please let me know a different width do you prefer by trying it locally.

kelvah:
> @kelvah FYI we are already using `widths` property. I set it to `width_33` just to be aligned with mock-ups.
I didn't realize it :+1: 
 
> This is how `width_66` looks like. Please let me know if you like it and I will commit. Otherwise, please let me know a different width do you prefer by trying it locally.

`width_66` is too big I think. I tried `width_50` and the URL is still cropped. Let's leave it as it is for now (`width_33`). Thank you.

vpellegrino:
> > 1. Then you would have a full-screen page instead of a drawer, and probably it will not be enough to show the full URL.
> > 2. You have a point. I'm not sure we should accomplish it in this PR, anyway we could trace it (if not already traced), WDYT @kelvah ?
> 
> 1. I've checked PF docs and it seems possible to increase the drawer width. See the [breakpoint example](https://www.patternfly.org/v4/components/drawer#breakpoint) where they set `<DrawerPanelContent widths={{ default: 'width_33' }}>`. I would give it a try.
> 2. I've missed it during my review. If it's not too much hassle I would do it here.

It seems that for accomplising the point 2), there are a few things to change. For example, mocked instances in `InstancesList` is not having the `published_at` prop, and the `DrawerContent`'s `panelContent` is expecting an instance details as soon as the page is loaded: instead we need to set the instance just when the user click on a specific row action.
So it is not an immediate change as expected.

vpellegrino:
@kelvah @manstis I added the drawer detail also the instances page.
Please let me know if the proposed solution is ok for you.

kelvah:
It works for me. Thanks @vpellegrino.
Quick question about `onDetailsClick`, is it a temporary solution? Do you plan to make row actions configurable at some point?

vpellegrino:
> It works for me. Thanks @vpellegrino. Quick question about `onDetailsClick`, is it a temporary solution? Do you plan to make row actions configurable at some point?

@kelvah yes, for now a callback seems enough from my point of view.
As soon as we have the delete in place, it will make sense to pass the whole actions configuration object, for sure.

jomarko (commented):
Thank you for the PR @vpellegrino .

- Am I correct this PR is just for bridge instances? processor instance details can not be shown?
- Was assumed a scenario of a quick switching instance details? Currently user need to close drawer, click on three dots, click on details. If the drawer would just shift original content, not overlay, going through details of multiple bridge instances would be faster from my point of view.
- Is it intentional `Status` is missing in the drawer?

vpellegrino:
> Thank you for the PR @vpellegrino .
> 
> * Am I correct this PR is just for bridge instances? processor instance details can not be shown?
> * Was assumed a scenario of a quick switching instance details? Currently user need to close drawer, click on three dots, click on details. If the drawer would just shift original content, not overlay, going through details of multiple bridge instances would be faster from my point of view.
> * Is it intentional `Status` is missing in the drawer?

Hi @jomarko 
Yes, this PR is just for bridge instance detail. Related to the other two questions, we followed the mock-up https://docs.google.com/presentation/d/1Uj4s7yUIVlvsY6kdISw05JDGTToptIOAeAUUol5Fwd0/edit?hl=en&forcehl=1#slide=id.g11820ca1acc_0_25 and I think that any proposal can be thrown during our UX interlock meetings.

jomarko:
@vpellegrino I am not attendee of such meetings. What is the way I put a question for its audience?

jomarko (approved):

vpellegrino:
> @vpellegrino I am not attendee of such meetings. What is the way I put a question for its audience?

I invited you

kelvah:
> Thank you for the PR @vpellegrino .
> 
> * Am I correct this PR is just for bridge instances? processor instance details can not be shown?
> * Was assumed a scenario of a quick switching instance details? Currently user need to close drawer, click on three dots, click on details. If the drawer would just shift original content, not overlay, going through details of multiple bridge instances would be faster from my point of view.
> * Is it intentional `Status` is missing in the drawer?

To answer your points @jomarko:

- processors don't need a drawer because they have a detail page with all their properties (reachable from the instance detail page, clicking on a processor name).
- we are aligning to a common design pattern within the console. so unfortunately not so much freedom for changes.
- same as above, but the content of the drawer will change (with permissions etc.). so this is only the first iteration. we could adjust things in the future.

PR #33, opened 2022-04-27
NO-JIRA: Renaming federated module and exposed component

kelvah:
Changing exposed component name to something more descriptive. Also making federate module name more readable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "smartevents-ui",
   "version": "0.0.1",
-  "federatedModuleName": "smartevents",
+  "federatedModuleName": "smart_events",
   "private": true,
   "scripts": {
     "prepare": "husky install",
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ npm install
 npm run start:federate
 ```
 
-This will run a dev server on http://localhost:9006 that will serve a federated module named `smartevents`.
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "smartevents-ui",
   "version": "0.0.1",
-  "federatedModuleName": "smartevents",
+  "federatedModuleName": "smart_events",
```
manstis, line 4: I wonder whether it would be better to make this consistent with [this](https://github.com/RedHatInsights/cloud-services-config/pull/1067/files#diff-c608be8e36b8dbaba0b0fc0d75c28bc58ab3a27167774c46f31f14319931c693R786)?

kelvah: Unfortunately I cannot use dash here. Federated modules plugin doesn't allow it and throws an error. Switched then to snake case.

manstis (approved):

package.json
jomarko: I s the readme fine without updates?
```
npm install
npm run start:federate

This will run a dev server on http://localhost:9006 that will serve a federated module named `smartevents`.
```

jomarko (commented):
One question as no expert in federated modules.

package.json
kelvah: good catch. fixed.

jomarko (approved):
Thank you @kelvah

PR #35, opened 2022-04-27
MGDOBR-643: Sync SDK and Openapi - 27-April-2022

jomarko:
For more details see:
- https://issues.redhat.com/browse/MGDOBR-643

PR #36, opened 2022-04-29
MGDOBR-471: Do not use inline responses

jomarko:
For more details see:
- https://issues.redhat.com/browse/MGDOBR-471

This PR contains also changes of
- #35 

This PR is dependent on
- https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/696


manstis (approved):
Great job @jomarko :+1: 

(I was just looking at your corresponding PR in `sandbox`).

It might be worth linking this PR from it (unless it's automatically done it) so the reviewers there can better understand (and appreciate) why we'd like the change made.

kelvah (approved):
Nice work @jomarko :ok_hand:

PR #40, opened 2022-05-04
MGDOBR-674: Plug "Processor detail" page to APIs

kelvah:
https://issues.redhat.com/browse/MGDOBR-674

Steps to test it:

- `npm run start:mocked`
- open example processors urls: [sink processor](http://localhost:8080/instance/3543edaa-1851-4ad7-96be-ebde7d20d717/processor/a72fb8e7-162b-4ae8-9672-f9f5b86fb3d7), [source processor](http://localhost:8080/instance/3543edaa-1851-4ad7-96be-ebde7d20d717/processor/f8f34af4-caed-11ec-9d64-0242ac120002)

PR #38, opened 2022-05-02
MGDOBR-654: [UI] OUIA in the 'Instance' component

sjamboro:
See: https://issues.redhat.com/browse/MGDOBR-654

PR #41, opened 2022-05-09
MGDOBR-668: Plug SE instances list to API

vpellegrino:
**Description:** https://issues.redhat.com/browse/MGDOBR-668
I started from the @kelvah branch (https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/40), so you will find a few commits related to his PR.
**In action:**
MGDOBR-668


kelvah (commented):
Thanks @vpellegrino, left you a couple of comments.

Not related to the changes in this PR, but could you please update the page size options in the toolbar to something like 10, 20, 50, 100?

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -193,34 +134,54 @@ const InstancesListPage = (): JSX.Element => {
...
+            totalRows={totalRows}
+            pageNumber={currentPage}
+            pageSize={currentPageSize}
+            onPaginationChange={(pageNumber, pageSize): void => {
+              const correctPageNumber =
+                pageSize === currentPageSize ? pageNumber : FIRST_PAGE;
+              setCurrentPage(correctPageNumber);
+              setCurrentPageSize(pageSize);
+              void getBridges(correctPageNumber, pageSize);
+            }}
```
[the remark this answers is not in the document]
vpellegrino: I am moving this block to a dedicated function.

I tested it a bit about the page size change near the subsequent polling execution (you can also try it locally). We have three cases:
1) Changing page size just before the polling execution will cause the polling to be executed with the current pagination parameter because we always perform the polling with the most updated version of `currentPage` and `currentPageSize`.
2) Changing page size after the polling execution will cause the "regular" call to be executed with the most updated version of `currentPage` and `currentPageSize`.
3) The "regular" call (caused by user action on pagination) and the polling calls are being executed simultaneously: only the last REST call will be completed. At the same time, the Axios Cancellation Token will cancel the old one. In any case, the "surviving" call will be executed with the updated pagination parameters.

cypress/integration/smoke/BasicTest.ts
```diff
@@ -29,14 +29,17 @@ describe("Basic Elements", () => {
...
-    cy.ouiaType("PF4/TableRow")
-      .should("have.length", 13)
-      .eq(1)
-      .find("td")
-      .eq(0)
-      .should("have.text", "Instance one")
-      .find("a[data-testid='tableInstances-linkInstance']")
-      .should("be.visible");
+  it("Mocked instances are visible", async () => {
+    cy.route("/bridges/*").as("getBridges");
```
sjamboro, line 33: @vpellegrino It seems that the `route` method might be deprecated in our cypress version (9.6.0) https://docs.cypress.io/api/commands/route. The `intercept` method has the same usage. See  https://docs.cypress.io/api/commands/intercept

vpellegrino: @sjamboro are you able to send edit suggestion or a pull request against my branch? Really appreciate. Thanks

src/app/components/TableWithPagination/TableWithPagination.tsx
```diff
@@ -81,18 +106,28 @@ export const TableWithPagination: FunctionComponent<
...
       </Toolbar>
-      <Table
-        actionResolver={actionResolver}
-        ariaLabel={tableLabel}
-        columns={columns}
-        cssClasses="overview__table"
-        rows={rows}
-      />
-      {getPagination(rows.length, true)}
+      {isLoading ? (
```
sjamboro, line 113: @vpellegrino there might be a merge conflict with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/38

cypress/integration/smoke/BasicTest.ts
vpellegrino: @sjamboro resolved with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/41/commits/9b0d8eeb7b5bb1fd384ecda6b4e8b1c823fa4429
Thanks for reviewing it!

sjamboro: @vpellegrino I am not sure if your solution runs correctly ... see 
image

There is a waiting rutine for a long time but the test is marked as succesfull.

sjamboro: @vpellegrino I am sorry for delay. I was running it localy and thinking how to fix it. I have created new JIRA https://issues.redhat.com/browse/MGDOBR-710.

vpellegrino: @sjamboro thank you for fixing it! Please, when you'll be ready involve me in the review of MGDOBR-710. I'm willing to learn how to wait for async calls in cypress

jomarko (changes requested):
### 01
@vpellegrino thank you for a PR. However *Description* is lost. I think we should remove this column completely. It is not present in backend at all.

Screenshot from 2022-05-11 07-38-13

jomarko (changes requested):
### 02
Pagination bottom right input box element is editable. We should:
- either do it read only
- or fix it to react on value change
Screenshot from 2022-05-11 07-46-11

jomarko (changes requested):
### 03
There is a case, when user is not informed anyhow, that pagination was clicked and some data are being fetched.

See my video https://issues.redhat.com/secure/attachment/12716752/pagination.webm

Once other page is selected:
- Screen should be cleared/emptied/greyed/loading popup should appear, or similar

What do you think please?

vpellegrino:
@jomarko thanks for your comments.
1/3 fixed
About point 2) please have a look at https://www.patternfly.org/v4/components/pagination/
We are using this PF component, so behavior you are observing is aligned with it
When you manually change the page number, without pressing **enter** nothing happens.
If you press **enter** the page is refreshed as well.
bottom_pagination
I would say that **if** this is a bug, it should be opened against PF. There is nothing we can do about it, since there is no possibility to configure a different behavior.

jomarko:
@vpellegrino thank you for
 1) fix - will recheck later
 2) explanation
 3) ???

vpellegrino:
> @vpellegrino thank you for
> 
> 1. fix - will recheck later
> 2. explanation
> 3. ???
As pointed in https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/41#issuecomment-1123235711,
`1/3 fixed`, so 1 **and** 3 points fixed.
In particular, for 3:
loading_pagination

jomarko (approved):
Thank you @vpellegrino 

@sthundat please notice, this PR will be in conflict with #43 . So if #41 is merged first. We need to update #43. Actually, I recommend to update #43 right now and merge it after #41.

sjamboro:
Similar for #38. It should be merge after this PR

sjamboro (approved):
LGTM ... InstanceListPage reflects changes in the `data.ts` file ... the pagination works

src/app/Instance/InstancesListPage/InstancesListPage.tsx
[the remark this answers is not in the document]
kelvah: Thanks for clarifying the three scenarios. Like you say the correct data is always shown. I'm going to approve.

I think there's only a minor rendering issue but it's not blocking for this PR. Also because it's not easy to reproduce :). So I'm not asking to fix it here.

When the pagination call is triggered right before the polling I think this happens:
1. pagination call starts
2. skeleton data is shown
3. polling request starts
4. pagination call gets canceled
5. skeleton data is hidden, previous data is shown
6. polling request succeeds
7. new items list with correct page/size is rendered on top of the previous one. here I could think the data was already fetched at step 5.

This is nitpicking of course. I also realize this is probably noticeable mainly with the mocked apis and their high response delay (1s). Requests to the real apis are usually faster, so the overlap will be less likely to happen.

kelvah (approved):
Thank you @vpellegrino :ok_hand:

src/app/Instance/InstancesListPage/InstancesListPage.tsx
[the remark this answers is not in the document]
kelvah: Yes i think it's missing skeleton due to call cancellation. Not so easy to reproduce. 
I've recorded it [here](https://issues.redhat.com/secure/attachment/12717112/pagination-followed-by-polling.webm)

src/app/Instance/InstancePage/InstancePage.tsx
```diff
@@ -230,6 +234,10 @@ const InstancePage = (): JSX.Element => {
               tableLabel={t(
                 "openbridgeTempDictionary:processor.processorsListTable"
               )}
+              onPaginationChange={(): void => {}} // @TODO MGDOBR-673
```
manstis, line 237: Sorry.. but what is "TODO"? The referenced JIRA is about plugging "...SE instance detail page to API" whereas these lines relate to pagination. Have I misunderstood something?

manstis (approved):
I don't see anything that has not already been asked and addressed. LGTM :+1: 

A minor comment about TODOs but nothing to stop merge.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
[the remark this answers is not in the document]
vpellegrino: I imagine that you can also reproduce with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/commit/514842147f70940f8f9486257f61e770db726495.
I tried it locally (a lot) and still couldn't reproduce it.
However, I think a _corner_ case happens when the polling is executed right before the cancelation, so the `isLoading` flag in the hook is set to `false` because, with polling, there is no loading skeleton.

src/app/Instance/InstancePage/InstancePage.tsx
vpellegrino: In the instance detail page, there is the list of processors belonging to it, represented by a table: that table is paginated.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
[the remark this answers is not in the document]
kelvah: Yes, I saw it with the latest version of the PR.
And yes my guess is the same as yours. The polling call that cancels the pagination call is also resetting the loading, I suppose [here](https://github.com/vpellegrino/sandbox-ui/blob/514842147f70940f8f9486257f61e770db726495/src/hooks/useBridgesApi/useGetBridgesApi.ts#L30) maybe.

Would you kindly open a jira to track it so we can investigate it later?

sthundat:
> Thank you @vpellegrino
> 
> @sthundat please notice, this PR will be in conflict with #43 . So if #41 is merged first. We need to update #43. Actually, I recommend to update #43 right now and merge it after #41.

sure @jomarko

PR #42, opened 2022-05-09
MGDOBR-675: Plug processors creation to APIs

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-675

This PR is dependent on #40

Steps to test it:

- `npm run start:mocked`
- open the create processor page: `http://localhost:8080/instance/3543edaa-1851-4ad7-96be-ebde7d20d717/create-processor`

While reviewing, you can filter only the last commit (60246e4) until #40 is merged.

### DIFF
--- a/src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
+++ b/src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
@@ -68,6 +68,8 @@ const CreateProcessorPage = (): JSX.Element => {
 
   useEffect(() => {
     if (processorError && axios.isAxiosError(processorError)) {
+      // TODO: replace error code string with a value coming from an error catalog
+      //  See https://issues.redhat.com/browse/MGDOBR-669 for more details.
       if (
         (processorError.response?.data as ResponseError).code === "OPENBRIDGE-1"
       ) {
--- a/src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
+++ b/src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import {
   PageSection,
   PageSectionVariants,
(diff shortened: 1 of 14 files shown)


src/types/Processor.ts
```diff
@@ -25,9 +25,40 @@ export interface EventFilter {
...
+
+export interface ProcessorValidation {
+  isValid: boolean | undefined;
+  errors: {
     [key: string]: string;
   };
 }
+
+export interface ConfigType {
+  value: string;
```
manstis, line 45: I found it a little confusing having this as `value` whereas it is checked against (type) `name`'s.

Just an opinion; no need to change it.

src/app/Processor/ProcessorEdit/useValidateConfigParams.ts
```diff
@@ -0,0 +1,98 @@
...
+          }
+          return fieldValidation.isValid;
+        }
+      }
+      return false;
+    },
+    [config, configTypes, resetValidation]
+  );
+
+  const validate = useCallback((): boolean => {
```
manstis, line 74: My aging brain would have appreciated some comments :rofl: 

e.g. "Collect separate validations for configured fields" etc.

Just an opinion, no need to change.

src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
```diff
@@ -7,32 +7,114 @@ import {
...
+    isLoading: isBridgeLoading,
+    error: bridgeError,
+  } = useGetBridgeApi();
+
+  useEffect(() => {
+    getBridge(instanceId);
+  }, [getBridge, instanceId]);
+
+  useEffect(() => {
+    if (bridgeError) {
```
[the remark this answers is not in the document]
kelvah: You're right, this is temporary. I think we should discuss a general error handling approach as soon as we finish the first round of api integration. Ideally in this case we should redirect to a "not found" page. But that's not the shared approach, I saw a different handling elsewhere.

src/types/Processor.ts
```diff
@@ -25,9 +25,40 @@ export interface EventFilter {
   value: string;
 }
 
-export interface BaseSource {
+export interface ProcessorRequestData {
+  name: string;
   type: string;
-  parameters: {
+  action: Action;
```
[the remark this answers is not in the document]
kelvah: :+1:
Fixed now.

src/app/Processor/ProcessorEdit/useValidateConfigParams.ts
kelvah: Sorry, I've rushed a bit this PR. I've added some comments to explain what's happening inside `useValidateConfigParams`.
Let me know if I should do the same elsewhere.

kelvah:
> I thought it might be handy to get a re-render when you click "Create" before submitting the request. If you create two processors, call the first `p1` and create it successfully; but that could be done as a separate JIRA:-
> 
> When creating the second call if `p1` as well. Complete the details. When you click create you'll rightly get told the name exists. Change it and onblur the error disappears. When you click "Create" the old error is shown and scrolled into view then the save completes.

Good catch. Totally missed it. It was caused by the error not being cleaned between different calls. Fixed now ([here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/1bcef29c9dc326b31489f185bacb341a0d8bdd7b/src/hooks/useProcessorsApi/useAddProcessorToBridgeApi.ts#L34)).

src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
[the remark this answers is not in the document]
kelvah: Created [MGDOBR-715](https://issues.redhat.com/browse/MGDOBR-715) to track the activity.

jomarko (changes requested):
I am not able to create processor. After data are submitted. It doesn't appear between processor instances.

https://issues.redhat.com/secure/attachment/12716763/processor.webm

manstis:
I agree with @jomarko comment. I didn't see new processors appear in the table - but wasn't sure I was expected to see them. They are being persisted somewhere as I am advised the name already exists if I try to create two with the same name.

kelvah:
> I am not able to create processor. After data are submitted. It doesn't appear between processor instances.

@jomarko @manstis, this PR is taking care of the processors creation page only. The processors list you are being redirected after creating a new processor is still showing hard-coded data, hence you are not seeing the updated processors list.

When you use the mocked version of the app, processors are actually saved inside the mocked service worker. That's why the "already existing name" check is working.

While we wait for the SE instance page to be plugged to APIs ([MGDOBR-673](https://issues.redhat.com/browse/MGDOBR-673)), you can manually check the updated list of processors by running the following fetch request in the browser developer tools console. Mind to replace `{bridgeId}` with the one you used creating the processor.
 
```
fetch("https://event-bridge-event-bridge-prod.apps.openbridge-dev.fdvn.p1.openshiftapps.com/api/v1/bridges/{bridgeId}/processors?page=0&size=10", {
  "method": "GET"
});
```

I realize it could have been easier to test this PR if we plugged pages in the natural navigation order (SE overview => Create SE instance => SE instance detail => Create Processor => and so on). We assigned the tasks in a different order to be able to work in parallel on them considering how the code is split and reused between pages.

vpellegrino:
@kelvah can you rebase this branch, please?

kelvah:
> @kelvah can you rebase this branch, please?

Done.

src/app/Processor/ProcessorEdit/ProcessorEdit.css
```diff
@@ -32,6 +32,10 @@
   padding: var(--processor-edit--spacing);
 }
 
+.processor-edit__form__notice {
+  margin-top: var(--pf-global--spacer--3xl);
```
vpellegrino, line 36: why do you need this custom margin top?

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
```diff
@@ -45,8 +52,81 @@ const ProcessorEdit = (props: ProcessorEditProps): JSX.Element => {
...
+      if (
+        request.transformationTemplate &&
+        request.transformationTemplate.trim().length > 0
+      ) {
+        requestData.transformationTemplate =
+          request.transformationTemplate.trim();
+      }
+      onSave(requestData);
+    }
+  };
```
vpellegrino, line 111: It seems that this function is performing too many actions: why don't you split it by multiple sub-functions?

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
```diff
@@ -45,8 +52,81 @@ const ProcessorEdit = (props: ProcessorEditProps): JSX.Element => {
...
+  useEffect(() => {
+    if (existingProcessorName) {
+      setIsSubmitted(true);
+      validate();
+    }
+  }, [existingProcessorName, validate]);
+
+  useEffect(() => {
+    if (isSubmitted) {
+      document.querySelector(".processor-field-error")?.scrollIntoView({
```
vpellegrino, line 122: I suggest dynamically putting (`processor-field-error`) in a React ref, and calling `scrollIntoView` on it, instead of querying the document.

vpellegrino (commented):
@kelvah thank you for this PR. I just left few comments

src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
```diff
@@ -7,32 +7,114 @@ import {
...
+  useEffect(() => {
+    if (addedProcessor) {
+      goToInstance();
+    }
+  }, [addedProcessor, goToInstance]);
+
+  useEffect(() => {
+    if (processorError && axios.isAxiosError(processorError)) {
+      if (
+        (processorError.response?.data as ResponseError).code === "OPENBRIDGE-1"
```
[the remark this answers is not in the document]
kelvah: I agree. I would try to avoid doing it manually if it's possible. I [opened a ticket](https://issues.redhat.com/browse/MGDOBR-669) a while back to set up scripts to automatically generate the error list like others SDKs do.

We could take care of it before doing [error handling consolidation](https://issues.redhat.com/browse/MGDOBR-715).

src/app/Processor/ProcessorEdit/ProcessorEdit.css
kelvah: I added it mainly for actions configuration. When an action is selected, the list of fields for its parameters is displayed.

With this margin there's some space between the end of the form and the end of the page and it's easier to understand there are some fields to take care of.

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
kelvah: The big chunk here is about a single task, preparing the request data to be sent to the APIs. The data from the form has more information than what's needed.

But yes, that's maybe too much for a `handleSubmit`, I've moved the request preparation to a separate function.

src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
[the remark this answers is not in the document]
vpellegrino: Probably it is worth to add a _TODO_ (with the ticket you're referring) for catching it faster in the future

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
kelvah: You caught me. The issue I was facing is that I'm not scrolling to a single specific node. I need to target the first input field in the page with validation issues, wherever it is. 

Doing it with refs means registering and un-registering the refs of fields with the order they appear on page, so to know which one to scroll to.

I took the shortcut of querying directly the document because it's way more easier, even if it's not the canonical react way of doing this things.

vpellegrino (approved):

src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx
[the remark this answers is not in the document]
kelvah: Of course. Done.

PR #45, opened 2022-05-12
[NO-JIRA]: Size issue on list of instances

vpellegrino:
The current returned `size` param from API does not contain pagination size, but length of returned elements.

### DIFF
--- a/mocked-api/handlers.ts
+++ b/mocked-api/handlers.ts
@@ -81,20 +81,22 @@ export const handlers = [
     const page = parseInt(req.url.searchParams.get("page") ?? "0");
     const size = parseInt(req.url.searchParams.get("size") ?? "10");
 
+    const items = db.bridge.findMany({
+        take: size,
+        skip: page * size,
+        orderBy: {
+          submitted_at: "desc",
+        },
+      });
+
     return res(
       ctx.status(200),
       ctx.delay(apiDelay),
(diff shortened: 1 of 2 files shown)


mocked-api/handlers.ts
```diff
@@ -259,27 +259,28 @@ export const handlers = [
...
-          .map((item) =>
-            cleanupProcessor(
-              item as unknown as Record<string | number | symbol, unknown>
-            )
-          ),
-        page: page,
-        size: size,
+        items,
+        page,
+        size: items.length,
```
kelvah, line 283: @vpellegrino Nothing wrong here because we also need this change for the "get processors" handler but you have to do it for "get bridges" too.

manstis: and errors (if you have a handler for it, IDK).

`Bridges`, `Processors` and `Errors` are the three resources in openapi that support pagination (at the API level).

manstis (approved):

mocked-api/handlers.ts
kelvah: We don't have an handler for Errors yet. I'm not sure we need it if we go with [MGDOBR-669](https://issues.redhat.com/browse/MGDOBR-669) first.

kelvah (approved):

PR #49, opened 2022-05-13
MGDOBR-673: Plug SE instance detail page to API

vpellegrino:
**Task:** https://issues.redhat.com/browse/MGDOBR-673
MGDOBR-673


cypress/integration/InstancesTest.ts
```diff
@@ -55,7 +55,7 @@ describe("Instances Test", () => {
 
   describe("Instance Page - Instance six", () => {
     beforeEach(() => {
-      cy.visit("/instance/d7e13602-b046-4120-b377-15d61e21c31");
+      cy.visit("/instance/3543edaa-1851-4ad7-96be-ebde7d20d717");
```
sjamboro, line 58: @vpellegrino the name of the test suite is "Instance six", I wanted to be sure that the details panel should contains different values than dummy data ("Instance one"). So feel free to check `d7e13602-b046-4120-b377-15d61e21c31a` or just rename the test suite `Instance Page - Instance one`.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -45,7 +45,7 @@ const InstancesListPage = (): JSX.Element => {
       accessor: "name",
       label: t("common.name"),
       formatter: (value: IRowData, row?: IRow): JSX.Element => {
-        const bridgeId = (row as Instance)?.id;
+        const bridgeId = (row as Instance)?.id ?? "";
```
manstis, line 48: Just for my information to understand TS better; why was this needed? And, is it a good idea? IIUC `bridgeId` could have been a `string` or `undefined`. If undefined the `Link` would be broken whereas using `""` gives a link that isn't broken. However either way represents an invalid state (it should never happen). No change needed; I just wanted to understand the reason behind this change.

sjamboro (changes requested):
LGTM ... just minor comment about the consistency of cypress tests.
Pagination works, details and processor list show actual mocked data.

@vpellegrino I am not sure if there should be some message when the list of processors is empty.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
sjamboro: @vpellegrino how can I simulate missing data? I am not able to delete them from the `data.ts` file.

vpellegrino: @manstis in the `BridgeResponse` interface, all fields are optional, so I made `Instance`'s field optional too.
Not a big issue, indeed. A broken link is properly handled in the detail page, by redirecting to the home (bridges list): https://github.com/vpellegrino/sandbox-ui/blob/6e3df9d8d9fa4086a427fcccc66fc897a559f866/src/app/Instance/InstancePage/InstancePage.tsx#L103

Moreover remember that this is a static check. Whenever API will return an empty string as a link, this is managed too.

vpellegrino: @sjamboro there (https://github.com/vpellegrino/sandbox-ui/blob/ef3de184493f3a4b4e540ab379a344d8f522114f/mocked-api/handlers.ts#L283) you can set empty array to fake empty processors.

cypress/integration/InstancesTest.ts
vpellegrino: Ok, renaming the test suite, thanks @sjamboro

vpellegrino:
> LGTM ... just minor comment about the consistency of cypress tests. Pagination works, details and processor list show actual mocked data.
> 
> @vpellegrino I am not sure if there should be some message when the list of processors is empty.

@sjamboro I added an empty state when no processors are available (@all feel free to suggest a better message).

src/types/Instance.ts
```diff
@@ -1,11 +1,10 @@
...
-  owner: string;
-  published_at: string;
-  status: string;
-  submitted_at: string;
+  id?: string;
+  endpoint?: string;
+  name?: string;
+  published_at?: string;
+  status?: string;
+  submitted_at?: string;
```
kelvah, line 9: This is just a curiosity. Is this interface still needed or could we use the generated one?

kelvah (approved):
Overall it looks good.

Should we remove the "Details" action from the processors table? I don't think we have a details drawer for them, do we?

src/types/Instance.ts
vpellegrino: Removed both interface and details action

vpellegrino:
Skipping since IT test with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/49/commits/83d11bd67805f0e4fdc37f264e425fbf5455dfa0 since it will be worked separately in https://issues.redhat.com/browse/MGDOBR-706

sjamboro (approved):
LGTM ... thank you @vpellegrino

PR #51, opened 2022-05-16
MGDOBR-706: Add assertation of the new RHOSE instance creation

sjamboro:
https://issues.redhat.com/browse/MGDOBR-706

I have changed ouiaId for a table row (from `id` to `name`). It allows me simpler constructions for Submit and Cancel tests.


vpellegrino (approved):
Looks good to me, thank you!

sthundat (approved):
LGTM

cypress/integration/InstancesTest.ts
```diff
@@ -12,41 +12,52 @@ describe("Instances Test", () => {
...
+          .should("be.visible")
+          .within(() => {
+            cy.ouiaId("new-name", "PF4/TextInput").type(newInstanceName);
+            cy.ouiaId("info-instance-available-soon", "PF4/Alert").should(
+              "have.text",
+              "Info alert:Your Smart Events instance will be ready for use shortly after creation."
+            );
+            cy.ouiaId("submit", "PF4/Button").click();
+          });
+        cy.wrap($modal, { timeout: 20000 }).should("not.exist");
```
kelvah, line 29: Why do we need to wait that long for the modal to close? The request to create an instance should be quick.

cypress/integration/InstancesTest.ts
```diff
@@ -12,41 +12,52 @@ describe("Instances Test", () => {
...
+          .should("be.visible")
+          .within(() => {
+            cy.ouiaId("new-name", "PF4/TextInput").type(newInstanceName);
+            cy.ouiaId("info-instance-available-soon", "PF4/Alert").should(
+              "have.text",
+              "Info alert:Your Smart Events instance will be ready for use shortly after creation."
+            );
+            cy.ouiaId("cancel", "PF4/Button").click();
+          });
+        cy.wrap($modal, { timeout: 7000 }).should("not.exist");
```
kelvah, line 58: Same question here. 7 seconds seems maybe too much? In this scenario the modal should close almost immediately.

src/app/components/Table/Table.tsx
```diff
@@ -99,7 +99,7 @@ export const Table: FunctionComponent<TableProps> = ({
         <Tbody>
           {transformRows(rows, columns).map((row: TableRow, rowIndex) => (
             <Tr
-              ouiaId={row.originalData?.id as string}
+              ouiaId={row.originalData?.name as string}
```
kelvah, line 102: It's not strictly related to this PR because the `id` was already used here before, but I'm not sure if it's a good idea to retrieve information from the original data. 

We are inside a generic Table component that shouldn't know what is the type of content rendered in the rows.

Probably it makes more sense to only use the `rowIndex` for both `ouiaId` and `key`, like `row_{rowIndex}`.

What do you think? Also asking @vpellegrino opinion on this.

kelvah (commented):
Thanks @sjamboro. I have some questions for you.

src/app/components/Table/Table.tsx
vpellegrino: Hi @kelvah
Thanks for asking
I originally added the usage of id for the react key property.
The reason was that all domain models have an id.
By the way I put on the key a fallback on row index.
I don't have a strong opinion. Your consideration about agnosticism from original data is totally fine.

cypress/integration/InstancesTest.ts
sjamboro: @kelvah the same problem sometimes happens that the test failed when the default timeout was just 4 seconds. I did not want to solve it when the UI is under development.

sjamboro: @kelvah what is your suggestion about timeouts? Do we want solve why 4 seconds is not enough?

src/app/components/Table/Table.tsx
sjamboro: @kelvah do you plan to use `Table.tsx` in filters? When do we start using ouiaId than such table's rows usually contain `id` or `name` as ouiaId. See https://github.com/kiegroup/kogito-apps/blob/main/ui-packages/packages/management-console/cypress/integration/processes/page-initial-load.ts#L78-L79

For me it is more readable and safer `cy.ouiaId("New instance", "PF4/TableRow").should("be.visible")` than `cy.ouiaType("PF4/TableRow").contains("New instance").should("be.visible")`. I have not got any performance comparison of these approaches. Both approaches are possible. The first approach compare the whole text and the name should be unique (Processor and Instance creation forms check it). The second one has a problem with "New instance one" and "New instance two". 

I have checked that you are using this table only with Processors and Instances. Both contains `id` and `name`.
image

When I create new Instance I know only the `name` (not `id`) and I am not sure if the order is stable. Do you mean that  `cy.ouiaId("index-0", "PF4/TableRow").should("have.text","New instance")` will be allways true?

You have totally rigth that curently we are sure that it is not a problem. But we should decide which data construction are applicable in  the `Table` component. Can we require `id` or `name` as mandatory fields? IMHO it is not a scope of this JIRA.

cypress/integration/InstancesTest.ts
sjamboro: @kelvah I am not sure about performance in GHA. Localy I have no problem with 4 sec.

kelvah: Thanks for the explanation @sjamboro. I tried to run the tests locally too and, like you say, they complete way faster.
I guess the problem is related to GHA having a limited amount of resources to run. So I guess we have to keep the high timeouts.

src/app/components/Table/Table.tsx
kelvah: > When I create new Instance I know only the `name` (not `id`) and I am not sure if the order is stable. Do you mean that `cy.ouiaId("index-0", "PF4/TableRow").find("td:first").should("have.text","New instance")` will be allways true?

It has to be true. Instances are sorted by default from the latest one. So the test could also be responsible to check that the default sorting order is working properly. And if we use the index as row key / ouiaId we could also avoid using `td:first`.

> 
> You have totally rigth that curently we are sure that it is not a problem. But we should decide which data construction are applicable in the `Table` component. Can we require `id` or `name` as mandatory fields? IMHO it is not a scope of this JIRA.

I agree about the scope, I mentioned it in my first comment. 

Things are technically working with the current solution. Like you pointed out, we have ids and names on both tables data. My comment was more about discussing if it's the right thing to do from a conceptual point of view.
The table should be agnostic about the rows content because it's a generic table. It's not an InstancesTable or ProcessorsTable.

Anyway, we can improve this in a separate ticket. I'm going to approve.

kelvah (approved):

cypress/integration/InstancesTest.ts
vpellegrino: Sorry for jumping into this conversation
It seems that in certain cases, high timeouts are not enough
image
https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/actions/runs/2370475727

I'm able to run IT locally with success (and fast)

How do we want to proceed @sjamboro @kelvah ?

PR #52, opened 2022-05-16
MGDOBR-677: Plug SE instance deletion to APIs

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-677

It's only possible to delete SE Instances with no processors inside at the moment.

So, to test this PR, you should try to delete instances with and without processors inside. It's possible to delete instances from the table in the home screen and from the "actions" dropdown inside the instance detail page.

If you run the app against the mocked APIs, "Instance one" is the only one with processors, while the others have none.


vpellegrino (approved):
The PR overall looks good to me.
Good job @kelvah

manstis (approved):
LGTM :+1:

kelvah:
@sjamboro we are missing your review :slightly_smiling_face: 
adding also @jomarko who maybe has some time to look at it.

jomarko (approved):
Thank you @kelvah

sjamboro (approved):
New functionality works without problems for common users. I reported here one situation when I forgot that I have already deleted the bridge instance. InstancePage is not pooling and I am able to see this page after deletion (copied address). I believe that it can be solved by different JIRA.

Feel free to merge this PR.

PR #54, opened 2022-05-19
MGDOBR-678: Plug "Edit Processor" page to APIs

vpellegrino:
Task: https://issues.redhat.com/browse/MGDOBR-678


mocked-api/handlers.ts
```diff
@@ -435,6 +435,120 @@ export const handlers = [
...
+          },
+          bridge: {
+            id: {
+              equals: bridgeId as string,
+            },
+          },
+        },
+      });
+
+      if (existingProcessors.length > 1) {
```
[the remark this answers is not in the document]
vpellegrino: Oh, it actually means that you renamed the edited processor like another one.
Does it make sense to you?

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
```diff
@@ -83,21 +112,87 @@ const ProcessorDetailPage = (): JSX.Element => {
...
+  const handleUpdateProcessorSaving = useCallback(
+    (processorRequest: ProcessorRequest) => {
+      if (
+        currentProcessor &&
+        processorNotChanged(currentProcessor, processorRequest)
+      ) {
+        setIsEditing(false);
+        return;
+      }
+      void updateProcessor(instanceId, processorId, processorRequest).then(() =>
```
manstis, line 150: Just asking.. this is a function invocation? Is it customary to include the return type in the invocation?

src/hooks/useProcessorsApi/useUpdateProcessorApi.ts
```diff
@@ -21,17 +20,23 @@ export function useUpdateProcessorApi(
 } {
   const [processor, setProcessor] = useState<ProcessorResponse>();
   const [error, setError] = useState<unknown>();
-  const [isLoading, setIsLoading] = useState(true);
+  const [isLoading, setIsLoading] = useState(false);
+  const auth = useAuth();
+
+  const getToken = useCallback(async (): Promise<string> => {
```
manstis, line 26: @kelvah has been moving the token and base URL into the hook. IDK if you want to do the same?

vpellegrino: And what is different from what is in place here? @manstis this is a hook

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
vpellegrino: This is a function exported from the hook. It is returning a promise of void.
On that promise I am listening to the success part in order to disable the edit mode when the PUT is executed.
Lint is not complaining, so I guess it is ok from type definition point of view

src/hooks/useProcessorsApi/useUpdateProcessorApi.ts
manstis: Since the hook's use is duplicated everywhere we use a custom "API" hook, @kelvah has been moving use of the hook into the custom "API" hook to avoid passing it as a parameter. The same for the baseConfig/URL. He has been moving both into the custom hook to avoid passing them as parameters.  For example [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/main/src/hooks/useBridgesApi/useCreateBridgeApi.ts).

manstis: LOL.... Unless that is exactly what you have done in this PR :D (I only just noticed the file name).

manstis (approved):
Thank-you @vpellegrino

mocked-api/handlers.ts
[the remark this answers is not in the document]
vpellegrino: @manstis fixed with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/54/commits/75e9bc5a72d0cc2c1985803718907f95e5d38b48
I just noticed something that could be interesting also for @kelvah 
When updating a processor, its status will not be "ready" anymore, so the status loaded on the detail page will not reflect the flow on the "backend".
Since we don't have a refresh button here, do you suggest considering a polling mechanism (as a future improvement)?

mocked-api/handlers.ts
```diff
@@ -626,7 +741,7 @@ const cleanupProcessor = (
   // removing properties not needed for the response
   const omitProperties = ["bridge"];
 
-  if (!(processor.filters as Array<Record<string, unknown>>).length) {
+  if (!(processor.filters as Array<Record<string, unknown>>)?.length) {
```
sjamboro, line 744: @vpellegrino It is right that I can not modify Filteres and Transformation

sjamboro: @vpellegrino Finally I was able to change Filteres but only if I changed the name of the procesor. Otherwise I get error `Processor with name 'Processor two' already exists` (in REST response ... not in UI). Is it correct?

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
```diff
@@ -121,33 +216,64 @@ const ProcessorDetailPage = (): JSX.Element => {
...
-                  dropdownItems={actionItems}
-                />
-              </SplitItem>
+              {!isEditing && (
+                <SplitItem>
+                  <Split hasGutter={true}>
+                    <SplitItem>
+                      <Button
+                        isAriaDisabled={
+                          currentProcessor.status !==
```
sjamboro, line 233: @vpellegrino Is it not possible to edit processor twice in the mock server? After the editation, the state is accepted and it is not changed to ready.

mocked-api/handlers.ts
sjamboro: > @vpellegrino Finally I was able to change Filteres but only if I changed the name of the procesor. Otherwise I get error `Processor with name 'Processor two' already exists` (in REST response ... not in UI). Is it correct?

@manstis could it be connected with your suggestion: `existingProcessors.length === 1 or existingProcessors.length > 0`?

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
vpellegrino: The state should transition to "ready" after few seconds. Then, it should be possible to edit the processor again

mocked-api/handlers.ts
vpellegrino: ok, looking into this handler, thank you @sjamboro

src/app/Processor/ProcessorEdit/ProcessorEdit.test.tsx
```diff
@@ -11,7 +11,12 @@ const setupProcessorEdit = (): {
   const onSave = jest.fn();
   const onCancel = jest.fn();
   const comp = customRender(
-    <ProcessorEdit onSave={onSave} onCancel={onCancel} isLoading={false} />
+    <ProcessorEdit
+      onSave={onSave}
+      onCancel={onCancel}
+      isLoading={false}
+      saveButtonLabel="Create"
+    />
```
sjamboro, line 19: @vpellegrino can you add tests here?

vpellegrino: @sjamboro to test what? This PR is about plugging API, that will be covered by e2e tests

mocked-api/handlers.ts
manstis: @vpellegrino It most certainly relates to [this](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/54/files/6cae153cea074c1510dbf58e73cf7899f86f5f7e#diff-3904e841d4e956c7e5bec8a10905e9049776263d0e5f502a3062121c42aa0d05R501). If you try to update a processor that already exists the count will be `> 0` (as there is `1` i.e. the one you're editing)... and yes it is caused by my proposal (that fixed the scenario I found, but broke it for @sjamboro's scenario). The problem is that it's OK for the name of a `Processor` "update" to be the same as an existing `Processor` **IF** it is the same `Processor` i.e. `updated.id = existing.id`. I think the check for existing processors needs to be a little smarter and check if the processor being updated is the same as the one that exists.

vpellegrino: ok @manstis thanks for the suggestion. I'm currently working on it

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
sjamboro: @vpellegrino Thank you, I tried to refresh InstancePage so I lost data. Currently I go back through the breadcrumb item and it works.

src/app/Processor/ProcessorEdit/ProcessorEdit.test.tsx
sjamboro: @vpellegrino you have added new `props` (processorTypeSection - https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/6cae153cea074c1510dbf58e73cf7899f86f5f7e/src/app/Processor/ProcessorEdit/ProcessorEdit.tsx#L188-L190). So I expect that it could affect the view of the form.

sjamboro:
@vpellegrino `test:it` fails because of this: 
image

There is not visible a page, it happen when there is a problem in the application.

vpellegrino:
> @vpellegrino `test:it` fails because of this: image
> 
> There is not visible a page, it happen when there is a problem in the application.

@sjamboro can you check if the failure there is the same of https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/runs/6553317229?check_suite_focus=true (your PR)?

mocked-api/handlers.ts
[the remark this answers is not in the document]
kelvah: @vpellegrino not sure about what's the approach we should follow here. I guess polling? Let's discuss it separately

src/app/Processor/ProcessorEdit/ProcessorEdit.test.tsx
vpellegrino: @sjamboro added https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/54/commits/8b673a16c596dbd0aadffe528cbce72e728d3267

sjamboro:
> > @vpellegrino `test:it` fails because of this: 
> > There is not visible a page, it happen when there is a problem in the application.
> 
> @sjamboro can you check if the failure there is the same of https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/runs/6553317229?check_suite_focus=true (your PR)?

@vpellegrino yes 
image

vpellegrino:
> > > @vpellegrino `test:it` fails because of this:
> > > There is not visible a page, it happen when there is a problem in the application.
> > 
> > 
> > @sjamboro can you check if the failure there is the same of https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/runs/6553317229?check_suite_focus=true (your PR)?
> 
> @vpellegrino yes image

Ok, so we can safely state that these failures are not related to this PR.
However, it seems that locally everything works fine.

mocked-api/handlers.ts
```diff
@@ -435,6 +435,124 @@ export const handlers = [
...
+              equals: bridgeId as string,
+            },
+          },
+        },
+      });
+
+      const processorNameCollision =
+        existingProcessors.length === 1 &&
+        existingProcessors[0].id !== processorId;
+      if (processorNameCollision || existingProcessors.length > 1) {
```
kelvah, line 504: Out of curiosity, is it possible that `existingProcessors  > 1`? If the query is on the name, which is unique, `existingProcessors` length can only be 1 or 0.

package.json
```diff
@@ -33,6 +33,8 @@
     "@rhoas/app-services-ui-components": "1.11.8",
     "@rhoas/app-services-ui-shared": "0.15.4",
     "keycloak-js": "9.0.0",
+    "lodash.isequal": "^4.5.0",
```
kelvah, line 36: :+1: 
You made me realize I've used `lodash` elsewhere as a transitive dependency. I'll have to [fix it](https://issues.redhat.com/browse/MGDOBR-757).

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
```diff
@@ -83,21 +112,87 @@ const ProcessorDetailPage = (): JSX.Element => {
       console.error(processorError);
       goToInstance();
     }
-  }, [bridgeError, processorError, goToHome, goToInstance]);
+    if (updateProcessorError) {
+      console.error(updateProcessorError);
```
kelvah, line 116: I think you should manage the "processor name already existing" scenario when the user is editing a processor. Similarly to [what was done here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/main/src/app/Processor/CreateProcessorPage/CreateProcessorPage.tsx#L74) during creation.

This is an error caused directly by the user, so we should take care of it in this phase.

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
```diff
@@ -83,21 +112,87 @@ const ProcessorDetailPage = (): JSX.Element => {
...
+      if (
+        currentProcessor &&
+        processorNotChanged(currentProcessor, processorRequest)
+      ) {
+        setIsEditing(false);
+        return;
+      }
+      void updateProcessor(instanceId, processorId, processorRequest).then(() =>
+        setIsEditing(false)
+      );
```
kelvah, line 152: Minor thing. I guess you could do `setIsEditing(false)` inside the `useEffect` that handles `updatedProcessor` (lines 102~104)

mocked-api/handlers.ts
vpellegrino: It is possible only if existing data are malformed. So, yes, it eventually could.

package.json
vpellegrino: Good to know

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
vpellegrino: Ok, I'll do

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
vpellegrino: Ok, then we can get rid of returning a Promise

vpellegrino:
@sjamboro @kelvah managed/answered all review feedback.

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
```diff
@@ -121,33 +230,65 @@ const ProcessorDetailPage = (): JSX.Element => {
...
-                      toggleIndicator={CaretDownIcon}
-                    >
-                      {t("common.actions")}
-                    </DropdownToggle>
-                  }
-                  isOpen={isActionsOpen}
-                  dropdownItems={actionItems}
-                />
-              </SplitItem>
+              {!isEditing && (
```
sjamboro, line 241: @vpellegrino the `ProcessorDetailPage.test.tsx` does not exist. Can you consider to create such file? There are several conditions which affect how the page will look like. I do not expexct that you will fix it right now.

sjamboro (approved):
LGTM ... approved I was able to:
- rename Processor, there was a warning about duplicity
- change other fields (filters, actions, etc) when the name was not changed
- edit it again and again (workaround for mocked app: via breadcrumb item)

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
vpellegrino: @sjamboro, it seems we don't have jest tests for pages. I expect them to be covered by IT.
At least, this is what we agreed til now.
Please, @kelvah, correct me if I'm wrong.

src/hooks/useProcessorsApi/useUpdateProcessorApi.ts
```diff
@@ -4,37 +4,42 @@ import {
...
+    return (await auth.kas.getToken()) || "";
+  }, [auth]);
+
+  const updateProcessor = (
     bridgeId: string,
     processorId: string,
     processorRequest: ProcessorRequest
-  ): Promise<void> => {
+  ): void => {
+    setIsLoading(true);
```
kelvah, line 35: You should probably clean the `error` and `processor` states when a new request starts.

Otherwise if a previous request failed, the error will still be present when running the new one.

This is causing the "existing name" error to briefly reappear when hitting the save button after fixing the name.

kelvah (approved):
LGTM. Thank you for the changes. :+1: 

Leaving another comment for your consideration.

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
kelvah: That's correct @vpellegrino

PR #55, opened 2022-05-19
MGDOBR-752: Provide links to the SE Instance detail page only when the Instance is "ready"

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-752

Steps to test it:
- create a new SE instance
- wait for it to appear in the list
- there should be no link on its name until it reaches the `ready` status

### DIFF
--- a/src/app/Instance/InstancesListPage/InstancesListPage.tsx
+++ b/src/app/Instance/InstancesListPage/InstancesListPage.tsx
@@ -31,7 +31,7 @@ import { TableWithPaginationSkeleton } from "@app/components/TableWithPagination
 import { useCreateBridgeApi } from "../../../hooks/useBridgesApi/useCreateBridgeApi";
 import axios from "axios";
 import { ResponseError } from "../../../types/Error";
-import { BridgeResponse } from "@openapi/generated";
+import { BridgeResponse, ManagedResourceStatus } from "@openapi/generated";
 
 const InstancesListPage = (): JSX.Element => {
   const { t } = useTranslation(["openbridgeTempDictionary"]);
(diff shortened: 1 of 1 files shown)


manstis (approved):
Super simple (and effective).

:superhero:

vpellegrino (approved):
Looks good to me

sjamboro (changes requested):
Hi @kelvah, can you put unit test(s) here? If there might be a problem accessing an instance before it is ready, we should check it.

The expectation of MGDOBR-752 was fulfilled.

On InstancePage, we could disable the "Create Procesor" button or the "Delete" action . I am not sure if a user has access to the ID of the new instance (via REST API).

sjamboro:
> > Hi @kelvah, can you put unit test(s) here? If there might be a problem accessing an instance before it is ready, we should check it.
> 
> @sjamboro I think we agreed to cover data-fetching components with E2E tests only.
> 
> I could agree this component has grown a bit too much and we could consider splitting it further between data fetching logic and templating.

I have created this issue https://issues.redhat.com/browse/MGDOBR-756. We can discuss the final location later. I understand that it is a nonsence to test it in both test suites (`test` and `test:it`). Currently both test suites are running in PR checks. IMHO the component test performs it quicker than a cypress test.

> 
> > On InstancePage, I am not sure if a user has access to the ID of the new instance (via REST API). Can he or she create own link? If yes, we could disable the "Create Procesor" button or the "Delete" action .
> 
> This means the user is manipulating the UI beyond what is allowed. In any case trying to create processors in this way will still result in a API failure. If we want to take care of this scenario, We could probably prevent direct access to the instance details page, maybe doing a redirection.

Such situation is not allowed but possible. The redirection sounds good. Maybe the simplier way is just show name and status plus some info warning about the situation. You can solve it separately.

> 
> > On InstanceListPage, is possible to delete an instance when the state is "Accepted" or "Provisioning"?
> 
> No, it's not possible. It's handled in my other PR #52
It is on my TODO list :) I will take a look at it

sjamboro (approved):
LGTM

PR #63, opened 2022-05-26
MGDOBR-676,677: Cover delete Processor and Instance

jomarko:
For more details see:
- https://issues.redhat.com/browse/MGDOBR-676
- https://issues.redhat.com/browse/MGDOBR-677

### DIFF
--- a/cypress/integration/DeleteTest.ts
+++ b/cypress/integration/DeleteTest.ts
@@ -1,0 +1,84 @@
+describe("Delete Test", () => {
+  it("Instance", () => {
+    cy.visit("/");
+
+    cy.ouiaId("Instances list table", "PF4/Table")
+      .ouiaId("Instance ten", "PF4/TableRow")
+      .find("td")
+      .then(($cells) => {
+        expect($cells).have.length(4);
+        expect($cells.eq(1)).have.text("ready");
+        cy.wrap($cells.eq(3)).ouiaType("PF4/Dropdown").click();
+      });
+
+    cy.ouiaType("PF4/DropdownItem").contains("Delete").click();
+
+    cy.ouiaId("delete-instance", "PF4/ModalContent").within(() => {
+      cy.ouiaId("delete-confirmation-value", "PF4/TextInput").type(
+        "Instance ten"
+      );
+      cy.ouiaId("confirm", "PF4/Button").click();
+    });
+
+    cy.ouiaId("delete-instance", "PF4/ModalContent").should("not.exist");
+
+    cy.ouiaId("Instances list table", "PF4/Table").within(() => {
+      // once delete confirmed, state should change
+      cy.ouiaId("Instance ten", "PF4/TableRow")
+        .find("td")
+        .eq(1)
+        .should("have.text", "deprovision");
+
+      // once deprovision is completed, entry should disappear
+      cy.ouiaId("Instance ten", "PF4/TableRow", { timeout: 30000 }).should(
+        "not.exist"
+      );
+    });
+  });
+
+  it("Processor", () => {
+    // Open 'Instance one'
+    cy.visit("/instance/3543edaa-1851-4ad7-96be-ebde7d20d717");
+
+    cy.ouiaId("instance-details", "PF4/Tabs")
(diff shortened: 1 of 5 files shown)


vpellegrino (approved):
The test structure is consistent and the code is easy to read.
Thank you @jomarko

manstis:
Sorry @jomarko I will not have time to review this before my PTO. 

I have removed myself as a reviewer.

cypress/integration/DeleteTest.ts
```diff
@@ -0,0 +1,85 @@
+describe("Delete Test", () => {
+  it("Instance", () => {
+    cy.visit("/");
+
+    cy.ouiaId("Instances list table", "PF4/Table")
+      .ouiaId("Instance ten", "PF4/TableRow")
+      .find("td")
+      .then(($cells) => {
+        expect($cells).have.length(4);
+        cy.wrap($cells.eq(3)).ouiaType("PF4/Dropdown").click();
```
sjamboro, line 10: Just information comment:
Maybe we should decide which approach we will use. Yours pure cypress solution or this approach https://github.com/jomarko/sandbox-ui/blob/MGDOBR-676%2C677/cypress/integration/InstancesTest.ts#L36-L37. Where I mix CSS selector.

I am not sure if there is necessary to assert the length. I never use it but the another pure cypress solution could be `.find("td").eq(3).ouiaType("PF4/Dropdown").click();`

cypress/integration/DeleteTest.ts
```diff
@@ -0,0 +1,85 @@
+describe("Delete Test", () => {
+  it("Instance", () => {
+    cy.visit("/");
+
+    cy.ouiaId("Instances list table", "PF4/Table")
+      .ouiaId("Instance ten", "PF4/TableRow")
+      .find("td")
+      .then(($cells) => {
+        expect($cells).have.length(4);
```
sjamboro, line 9: To be consistent with Processor, can you check status is ready

sjamboro (approved):
Do you plan to add the assertation of:

- the validation
- the error messages

kelvah (approved):
LGTM

cypress/integration/DeleteTest.ts
jomarko: sure, done

cypress/integration/DeleteTest.ts
jomarko: I do not think there is a simple answer what approach is the best. Speaking about case that `get`/`find` returned collection of elements.

I think it depends if we are interested in single element of returned collection vs in multiple elements of given collection.

For the first case I use , after your review comments, approach like this:
```
...
        .find("td")
        .eq(1)
        .should("have.text", "deprovision");
```

for the second case:
```
...
        .find("td")
        .then(($cells) => {
          expect($cells).have.length(6);
          expect($cells.eq(0)).have.text("Processor one");
          expect($cells.eq(3)).have.text("ready");
          cy.wrap($cells.eq(5)).ouiaType("PF4/Dropdown").click();
        });
```

jomarko:
@sjamboro 
```
Do you plan to add the assertation of:

the name validation (insert a wrong name of deleted instance)
the error messages (processors exist)
```

Not as part of this PR, please create a jira for me if you think such tests are needed. Especially insert wrong name of deleted instance - I believe it is feature of patternfly, that we shouldn't test.

@vpellegrino ? ^

vpellegrino:
> @sjamboro
> 
> ```
> Do you plan to add the assertation of:
> 
> the name validation (insert a wrong name of deleted instance)
> the error messages (processors exist)
> ```
> 
> Not as part of this PR, please create a jira for me if you think such tests are needed. Especially insert wrong name of deleted instance - I believe it is feature of patternfly, that we shouldn't test.
> 
> @vpellegrino ? ^

@jomarko I checked and it seems that the logic for delete confirmation is managed on our side (https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/4e3a4108fe496954f545d86ef3135975d92ea310/src/app/components/DeleteModal/DeleteModal.tsx#L57)
However, I agree to manage these tests in a separate task.

PR #61, opened 2022-05-24
MGDOBR-742: Set up config context and update auth context

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-742

PR #56, opened 2022-05-23
MGDOBR-727: Update PatternFly to version 2022.04 (and related deps)

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-727

Updating PF to 2022.04 and related packages.

No relevant visual changes are expected following the update. 
The [`FormSelect` bug](https://github.com/patternfly/patternfly/issues/4772#event-6430067183) we reported was kindly fixed but  it got released with 2022.05. So we have to wait a bit more.

PR #64, opened 2022-05-27
NO-JIRA: Wait instances are loaded before collapsing panel

jomarko:
Click on panel toggle before data are loaded causing panel is not collapsed

On main the fail looks like:
Screenshot from 2022-05-27 08-21-56

It can be spotted aslo on PRs:
- #61 https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/runs/6613185257?check_suite_focus=true

### DIFF
--- a/cypress/integration/smoke/BasicTest.ts
+++ b/cypress/integration/smoke/BasicTest.ts
@@ -16,13 +16,33 @@ describe("Basic Elements", () => {
     cy.ouiaId("smart-events", "PF4/NavItem").then(($item) => {
       //toggle menu side bar - both directions
       if ($item.is(":visible")) {
+        // wait until data are loaded
+        cy.ouiaType("PF4/TableRow")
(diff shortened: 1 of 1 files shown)


sjamboro:
@jomarko here is the screenshot of the failure directly from MGDOBR-742. 
image

sjamboro:
@jomarko @kelvah do you think that it might be connected with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/56.

I have rebase MGDOBR-723 and this test starts failing too.

jomarko:
Yes, it may be related. But in this case, I do not think we can fix the failing test by revetrting PF upgrade. We need somehow stabilize the test, or remove it.

jomarko:
Test fixed as part of https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/53

PR #73, opened 2022-06-02
MGDOBR-637: Align UI TS SDK with rhose-api.yaml

jomarko:
For more details please see:
- https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/816
- https://issues.redhat.com/browse/MGDOBR-637


scripts/openapi-sdk/generate_openapi_client.sh
```diff
@@ -27,4 +27,5 @@ npx --yes @openapitools/openapi-generator-cli generate -g typescript-axios -i \
 "$openapi_filename" -o "$output_path" \
 --remove-operation-id-prefix \
 --package-name=$package_name \
---additional-properties=$additional_properties
+--additional-properties=$additional_properties \
+--type-mappings object=Object
```
jomarko, line 31: This change is important, so we get:
```
export interface Action {
    /**
     * 
     * @type {string}
     * @memberof Action
     */
    'type': string;
    /**
     * 
     * @type {{ [key: string]: Object; }}
     * @memberof Action
     */
    'parameters': { [key: string]: Object; };
}
```

instead of:
```
export interface Action {
    /**
     * 
     * @type {string}
     * @memberof Action
     */
    'type': string;
    /**
     * 
     * @type {{ [key: string]: object; }}
     * @memberof Action
     */
    'parameters': { [key: string]: object; };
}
```

Please notice `Object` vs `object`. This particular class, in java has field declaration like `Map<String, Object> parameters;`. So values in java can be strings. In typescript, `object` would force just non primitive values. String wouldn't be allowed.

openapi/generated/api/processors-api.ts
```diff
@@ -162,12 +166,15 @@ export const ProcessorsApiAxiosParamCreator = function (configuration?: Configur
...
+         * @param {string} [name] 
          * @param {number} [page] 
          * @param {number} [size] 
+         * @param {Set<ManagedResourceStatus>} [status] 
+         * @param {ProcessorType} [type] 
          * @param {*} [options] Override http request option.
          * @throws {RequiredError}
          */
-        listProcessors: async (bridgeId: string, page?: number, size?: number, options: AxiosRequestConfig = {}): Promise<RequestArgs> => {
+        listProcessors: async (bridgeId: string, name?: string, page?: number, size?: number, status?: Set<ManagedResourceStatus>, type?: ProcessorType, options: AxiosRequestConfig = {}): Promise<RequestArgs> => {
```
jomarko, line 177: If we check, coresponding java class, there are parameters `@NotEmpty @PathParam("bridgeId") String bridgeId, @Valid @BeanParam QueryProcessorResourceInfo queryInfo` 

I am not sure if that is an issue of code generation, that `QueryProcessorResourceInfo` is extracted into particular parameters. What do you think?

src/app/Processor/ProcessorEdit/SourceEdit/SourceEdit.tsx
```diff
@@ -171,7 +171,7 @@ const SourceEdit = (props: SourceEditProps): JSX.Element => {
                     aria-describedby={field.name}
                     isRequired={true}
                     isDisabled={isDisabled}
-                    value={parameters[field.name] ?? ""}
+                    value={(parameters[field.name] as string) ?? ""}
```
jomarko, line 174: same question as above

src/app/Processor/ProcessorEdit/useValidateConfigParams.ts
```diff
@@ -59,7 +59,9 @@ export function useValidateConfigParams(
         if (field) {
           // Run the field validation function provided by configTypes with the current
           // value of the field
-          const fieldValidation = field.validate(config.parameters[fieldName]);
+          const fieldValidation = field.validate(
+            config.parameters[fieldName] as string
+          );
```
jomarko, line 64: where do we have validate implementation please? I think we need now in that implementation check what is the type of the passed parameter? is that actually implemented on backend?

jomarko (commented):
Please notice, all existing usages of:
- `listProcessors`
- `getProcessors`
- `getBridges`

I just tried to get somehow green build, however I need an assistance, guidance, how to handle the changes in the API. Currently I tried just to pass `undefined` for all new optional method parameters.

scripts/openapi-sdk/generate_openapi_client.sh
kelvah: I see `Object` is avoiding us a lot of trouble here but I think we should avoid it. It's usually discouraged to use it as a type.

We could stay with `{ [key: string]: object; }`. But honestly I think it's harder to manage than having just `parameters: object`.

Ideally it should be like `{ [key: string]: unknown; }`, which is the closest to the actual thing inside `parameters`. but I'm not sure we have a way to generate it. So maybe we could just leave `parameters: object` and do type assertions where needed.

Also, taking a look around, many things that need to be fixed following the change to `parameters` could be removed after we switch to the dynamically generated forms (`ActionEdit`, `ProcessorDetail`, the validation hook etc.). 

So we have also the option to keep this PR on hold and update the SDK together with the changes to the processor action/source management.

openapi/generated/api/processors-api.ts
kelvah: I think everything is fine from the API perspective. `name`, `status`, `type` are all new optional filters that it's possible to pass as query params to get the processors list

scripts/openapi-sdk/generate_openapi_client.sh
jomarko: _So we have also the option to keep this PR on hold and update the SDK together with the changes to the processor action/source management._ what is the estimate for it @kelvah please? Or in other words, what jira blocks this?

kelvah: There are no blockers at the moment. I've started to look into it. Difficult to do an estimation though. I need some more time. Let's chat about it.

vpellegrino:
@jomarko about your question and type issue, I opened a PR against your branch: https://github.com/jomarko/sandbox-ui/pull/2 with a description for it. Please let me know if you have further questions.
`lint`, `format`, and `test` scripts are green in my local env

jomarko:
@vpellegrino thank you for a guidance, how to cast. I changed `{[key: string]: string}` to `{[key: string]: unknown}` as that was suggested by @kelvah  I believe. 

Would you mind to re-review?

src/hooks/useProcessorsApi/useGetProcessorsApi.ts
```diff
@@ -10,8 +12,10 @@ import { useSmartEvents } from "@contexts/SmartEventsContext";
 export function useGetProcessorsApi(): {
   getProcessors: (
     bridgeId: string,
+    filterName?: string,
```
kelvah, line 15: See my comment inside `useGetBridgesApi.ts`

kelvah (approved):

src/hooks/useBridgesApi/useGetBridgesApi.ts
```diff
@@ -36,7 +37,7 @@ export function useGetBridgesApi(): {
         })
       );
       bridgeApi
-        .getBridges(pageReq, sizeReq, {
+        .getBridges("", pageReq, sizeReq, new Set<ManagedResourceStatus>(), {
```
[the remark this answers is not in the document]
jomarko: yes, I unified the approach now, thank you for spotting @kelvah

PR #86, opened 2022-06-30
MGDOBR-790: [UI] E2E tests - Create PoC

sjamboro:
https://issues.redhat.com/browse/MGDOBR-790

I have created a simple test which send data via REST API and then check if they are present in UI.

All neccessary steps should be mentioned in README.

This test is not a part of GHA (right now) but it is possible to run it locally.

### DIFF
--- a/cypress/README.md
+++ b/cypress/README.md
@@ -38,11 +38,11 @@ npm run start
 Get a valid offline [token](https://console.redhat.com/openshift/token).
 
 ```
+export CYPRESS_USER=<replce with your value>
+export CYPRESS_PASSWORD=<replce with your value>
 export OPENSHIFT_OFFLINE_TOKEN=<REPLACE WITH YOUR TOKEN>
 export CYPRESS_OB_TOKEN="$(curl -s --insecure -X POST https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token --header 'Content-Type: application/x-www-form-urlencoded' --data-urlencode 'client_id=cloud-services' --data-urlencode 'grant_type=refresh_token' --data-urlencode "refresh_token=$OPENSHIFT_OFFLINE_TOKEN" | jq --raw-output '.access_token')"
 ```
-
-Set `USER` and `PASSWORD` in [the configuration](cypress.e2e.json).
 
 If the test suite fail as Unauthorized than perform again:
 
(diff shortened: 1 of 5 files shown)


cypress/README.md
```diff
@@ -20,4 +20,73 @@ npm run test:it
...
+
+```
+
+**Run E2E tests**
+
+```
+npm run cypress:open:e2e
+```
+
+Clear entities which were added by the test suite.
```
jomarko, line 60: I think we can implement this quite easily with code like bellow?
```
  after(() => {
    const token = Cypress.env("OB_TOKEN");
    const authorization = `Bearer ${token}`;
    const options = {
      method: "DELETE",
      url: `"https://event-bridge-event-bridge-prod.apps.openbridge-dev.fdvn.p1.openshiftapps.com/api/v1/bridges/" + ${bridgeId}`,
      headers: {
        "Content-Type": "application/json",
        Accept: "application/json",
        Authorization: authorization,
      },
    };
    cy.request(options).then((response) => {
      expect(response.body).to.not.be.null;
      expect(response.isOkStatusCode).to.equal(true);
    });
  });
```

?

jomarko (commented):
Thank you @sjamboro . 

I have tried manually the first part of readme and it works!

It is huge achievement! I have two inline comments.

sjamboro:
@jomarko Can you take a look at my changes?

sjamboro:
@kelvah Can you review this PR?

jomarko (approved):
Thank you @sjamboro !

cypress/README.md
```diff
@@ -20,4 +20,73 @@ npm run test:it
...
+
+## Run UI with real data locally
```
kelvah, line 62: Do we want to document this within the repo?

Asking because the whole process is hackish.

We already have mocked APIs + `dev` APIs to test new features. Do we need another way of doing this?

kelvah (commented):
Thanks @sjamboro. I have a question.

cypress/README.md
jomarko: well it is good question, I can imagine this can be beneficial for testing latest features of backend locally, if you mind we have this documented publicly, maybe we can move it ot some google document, or redhat source page?

sjamboro: @kelvah I have doubted to paste or delete this chapter here. As you mention it is `It is not the supported usecase.` I would like to mention it somewhere, because it could be useful. I have no problem to delete it.

kelvah: Thanks @sjamboro 

@jomarko +1 on moving this instructions into a google document.

kelvah (approved):

PR #89, opened 2022-07-05
MGDOBR-906: Verify new Google PubSub sink action

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-906

Checking the new PubSub sink.

There's a potential issue with the `gcp_service_account_key` property. It seems to be a credential type of field but it doesn't have the `password` format. Logged the issue inside [MGDOBR-901
](https://issues.redhat.com/browse/MGDOBR-901)

Just updating the corresponding mock for now.

### DIFF
--- a/mocked-api/schemasData.ts
+++ b/mocked-api/schemasData.ts
@@ -45,7 +45,7 @@ export const schemaCatalogData = [
     name: "Google PubSub",
     description: "Send the event to Google PubSub",
     type: "action",
-    href: "/api/v1/schemas/actions/google_pubsub_sink_0.1",
+    href: "/api/smartevents_mgmt/v1/schemas/actions/google_pubsub_sink_0.1",
   },
   {
     kind: "ProcessorSchemaEntry",
--- a/mocked-api/schemasData.ts
+++ b/mocked-api/schemasData.ts
@@ -41,6 +41,14 @@ export const schemaCatalogData = [
   },
   {
     kind: "ProcessorSchemaEntry",
+    id: "google_pubsub_sink_0.1",
+    name: "Google PubSub",
+    description: "Send the event to Google PubSub",
+    type: "action",
+    href: "/api/v1/schemas/actions/google_pubsub_sink_0.1",
+  },
+  {
+    kind: "ProcessorSchemaEntry",
     id: "aws_s3_source_0.1",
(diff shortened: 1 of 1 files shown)


mocked-api/schemasData.ts
```diff
@@ -318,6 +326,36 @@ export const schemasData: { [key: string]: object } = {
...
+        description:
+          "The Destination Name. For the consumer this will be the subscription name, while for the producer this will be the topic name.",
+        type: "string",
+      },
+      gcp_service_account_key: {
+        title: "Service Account Key",
+        description:
+          "The Service account key that can be used as credentials for the PubSub publisher/subscriber in base64 encoding.",
+        type: "string",
+        "x-group": "credentials",
```
manstis, line 354: This is an interesting one. The back-end looks for a field of format `password` and not this `x-group: credentials` property. Meaning that the front-end considers it _secret_ whereas the back-end does not. However since "updates" to _secret_ fields uses the magic "empty object" pattern it is unlikely to affect operation.

@kostola FYI. It is likely [this](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/863) will show this field in the response as the [schema](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/processor/configuration/src/main/resources/schemas/actions/google_pubsub_sink_0.1.json) does not have a password field.

vpellegrino (approved):
Looks good. Thank you

mocked-api/schemasData.ts
```diff
@@ -39,6 +39,14 @@ export const schemaCatalogData = [
     type: "action",
     href: "/api/v1/schemas/actions/aws_lambda_sink_0.1",
   },
+  {
+    kind: "ProcessorSchemaEntry",
+    id: "google_pubsub_sink_0.1",
+    name: "Google PubSub",
+    description: "Send the event to Google PubSub",
+    type: "action",
+    href: "/api/v1/schemas/actions/google_pubsub_sink_0.1",
```
jomarko, line 48: please update the href to be aligned with latest main

kelvah: Done.

jomarko (approved):
I do not have access to any running Google PubSub. Checked just the sandbox-ui form.

PR #91, opened 2022-07-06
MGDOBR-898: Public API returns ErrorsResponse

vpellegrino:
**Task**: https://issues.redhat.com/browse/MGDOBR-898
**Description**: Changing the function `getErrorCode` in `errorHelpers` to consider the first error returned by the APIs.
Consider that we could add new functions to the helper when we need to report multiple errors on the page or search for a particular error code among all returned error items.

### DIFF
--- a/mocked-api/handlers.ts
+++ b/mocked-api/handlers.ts
@@ -118,10 +118,15 @@ export const handlers = [
       ctx.status(404),
       ctx.delay(apiDelay),
       ctx.json({
-        ...error_not_found,
-        reason: `Bridge with id '${
-          bridgeId as string
-        }' for customer 'XXXXXXXX' does not exist`,
+        kind: "ErrorList",
+        items: [
+          {
+            ...error_not_found,
+            reason: `Bridge with id '${
+              bridgeId as string
+            }' for customer 'XXXXXXXX' does not exist`,
+          },
+        ],
       })
     );
   }),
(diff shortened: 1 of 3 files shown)


mocked-api/handlers.ts
```diff
@@ -118,10 +118,15 @@ export const handlers = [
       ctx.status(404),
       ctx.delay(apiDelay),
       ctx.json({
-        ...error_not_found,
-        reason: `Bridge with id '${
-          bridgeId as string
-        }' for customer 'XXXXXXXX' does not exist`,
+        kind: "ErrorList",
```
manstis, line 121: I'd have expected this to be `ErrorsResponse` not `ErrorList`.

However it is not _used_ anywhere so doesn't strictly matter.

vpellegrino: Sorry, I took it from a [Jira comment](https://issues.redhat.com/browse/MGDOBR-704):
image
I'm going to update it immediately

openapi/generated/errorHelpers.ts
```diff
@@ -7,15 +8,17 @@ import { AxiosError } from "axios";
...
- * Get the error code from the API error
+ * Get the error code related to the first error item returned from the API
  *
  * @param error generic error returned from function
  * @returns error code (one of fields of APIErrorCodes)
  */
 export const getErrorCode = (error: unknown): string | undefined => {
-    return (error as AxiosError).response?.data?.code;
+    const errorListResponse = (error as AxiosError).response?.data as ErrorListResponse;
+    return errorListResponse?.items?.length ? errorListResponse?.items[0].code : undefined;
```
jomarko, line 23: as this is part of `openapi/generated` folder, won't we lost this change with every sdk re-generation?

jomarko (commented):
Thank you for a PR @vpellegrino . One question inline.

kelvah (approved):

openapi/generated/errorHelpers.ts
vpellegrino: Only `model` and `api` folders are going to be removed during (re)generation: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/2a3e1408376ea5847113088eabef61889332536c/scripts/openapi-sdk/generate_openapi_client.sh#L20

jomarko (approved):

PR #95, opened 2022-07-12
[NO-JIRA] Format just staged files

jomarko:
For more details see:
- https://prettier.io/docs/en/precommit.html#option-6-shell-script


kelvah (approved):

sjamboro (approved):
LGTM ... thank you for a quick fix

PR #97, opened 2022-07-13
MGDOBR-937: Verify new Google PubSub source processor

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-937

Hopefully this is the last one!

Form generated from schema works fine. Needed changes are logged inside [MGDOBR-901](https://issues.redhat.com/browse/MGDOBR-901). For now it's just the Service Account Key missing the proper password definition.

### DIFF
--- a/mocked-api/schemasData.ts
+++ b/mocked-api/schemasData.ts
@@ -79,6 +79,14 @@ export const schemaCatalogData = [
     type: "source",
     href: "/api/smartevents_mgmt/v1/schemas/sources/slack_source_0.1",
   },
+  {
+    kind: "ProcessorSchemaEntry",
+    id: "google_pubsub_source_0.1",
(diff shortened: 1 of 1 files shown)


vpellegrino (approved):

manstis (approved):
LGTM :+1:

kelvah:
> One observation: there is possible to set negative integers which is nonsense but I assume that it depends on the source definition and UI need to support full interger when the definition defines it.

Agree, that should be defined in the schema. PRs like this one are mainly intended to do a formal check about all the schemas that were introduced over time, verifying that the form is generated correctly from the json-schema definition.

We'll take care of improvements/changes to the schemas with MGDOBR-901. 
I've added a note there proposing to add extra validation for positive integers on "Max Messages Per Poll" and "Concurrent Consumers"

PR #98, opened 2022-07-13
MGDOBR-806: Manage malformed transformation template error

vpellegrino:
**Task**: https://issues.redhat.com/browse/MGDOBR-806
image


manstis (commented):
Before I review completely I am a little underwhelmed.

Transformations was the first (and currently only) field that can return multiple errors and we did quite a bit of work on the backend to ensure the individual parsing errors were available to the UI. 

A quick skim of this PR suggests returning a _single_ `ERROR_22` would have been enough. Is this how it's to work? If so I will simply review as it is; or did you want to show all of the potential errors?

vpellegrino:
> 

@manstis, I am sorry to hear that and for falling short of your expectations.
Maybe I misunderstood the goal of https://issues.redhat.com/browse/MGDOBR-806.
If anything is missing from a design point of view, could you please add a comment to it or change the description accordingly? I would be happy to work on improving this task before moving to a new one.
Maybe there is a dedicated task to manage multiple errors for this field, and I am unaware of it.
@kelvah, do you have any feedback about it?

kelvah:
Hi @vpellegrino, there isn't another ticket for multiple errors AFAIK. I think we should manage them with this PR. The jira description is a bit vague about this, it was written long ago when the validation errors where quite different. Sorry for the misunderstanding. 

We should display all the errors related to the transformation template below it. Also because they carry relevant information like the row number and the specifics of the errors. They are useful to provide a fix.

kelvah:
Just to clarify, we will have to take care of multiple errors on different fields (like processor name and transformation template). But the APIs do not work this way ATM and they will return only one type of error at the time.

But, we could receive multiple errors of the same type about the transformation template. This is already supported, so we can already start covering it.

vpellegrino:
Temporarily closing this PR since it seems there is something missing in the expected result.

manstis:
I am so sorry @vpellegrino and I apologise if I was a bit abrupt in my first reply. Please ping me if you'd like an example of what the backend produces for multiple transformation errors. Happy to help.

vpellegrino:
As agreed with @manstis, reopening this PR

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
```diff
@@ -404,14 +411,33 @@ const ProcessorEdit = (props: ProcessorEditProps): JSX.Element => {
...
+                                    ? "processor-field-error"
+                                    : ""
+                                )}
                                 height={"300px"}
                                 isLineNumbersVisible={true}
                                 code={transformation}
-                                onChange={setTransformation}
+                                onChange={(value): void => {
+                                  setTransformation(value);
+                                  resetTransformationStatus?.();
```
kelvah, line 425: I was about to start reviewing but then I stopped to think about this, clearing the TT error on template change.

TL;DR: I think we should probably avoid it.

I did a similar thing for the "already existing processor name" error. As soon as the user changes the name, the error is gone. I still have mixed feelings from time to time when I see this behavior and I question it in my mind. 

When changing the name we are sure that the error about the previous name will be gone. Still, the user could pick another name already taken (how unlucky), triggering again the error after submitting.

We could say this is an "optimistic approach". In its defense we know that once the name is changed, you can try again and take another shot with a chance of picking a valid one.

About TT, the situation is similar but there's a difference. When changing the TT, we don't know if the change has any chance of solving the issue. Maybe the problem was caused by a missing bracket. If I just rename a variable instead, there's no chance the error will go away. It's difficult then to keep the same "optimistic approach".

I know this could look inconsistent with the "resource name validation" treatment. I am not sure if the difference between the two is big enough to justify two different strategies. Anyway, we could also decide to remove the _clean-on-change_  behavior from the processor/bridge name too.

@vpellegrino @manstis What do you think? I'm curious to hear from you.

manstis:
@vpellegrino Did I read somewhere you was going to update this to include the parser message and not just use a generic message? If so, could you please ping me again when this is ready for review. Thank-you. If not, please also ping me so I know this needs to be looked at!!

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
vpellegrino: Thanks for raising this interesting point @kelvah 
I reasoned about it while developing, and I concluded that from a user perspective, the validation should be consistent for all fields, except if there is a valid reason.
However, for this field, a new error will likely be raised after clicking on submit, so I will avoid clearing the TT error on template change.

vpellegrino:
@manstis @kelvah I accomplished your suggestions. PR is ready to be reviewed.

manstis (approved):
LGTM :+1:

kelvah (approved):
LGTM

jomarko:
Handing over the review to @sjamboro

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
```diff
@@ -215,6 +219,14 @@ const ProcessorDetailPage = (): JSX.Element => {
         actionModalMessage.current = t(
           "processor.errors.cantUpdateProcessorBecauseNotReadyState"
         );
+      } else if (
+        isServiceApiError(updateProcessorError) &&
+        getErrorCode(updateProcessorError) === APIErrorCodes.ERROR_22
+      ) {
+        setMalformedTemplate(
+          getErrorReason(updateProcessorError) ??
```
sjamboro, line 227: I do not know how display malformedTemplate if I am not able to create a processor with malformed Template.

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
```diff
@@ -66,6 +69,7 @@ export interface ProcessorEditProps {
 const ProcessorEdit = (props: ProcessorEditProps): JSX.Element => {
   const {
     existingProcessorName,
+    malformedTransformationTemplate,
```
sjamboro, line 72: The new parameter in ProcessorEdit, it could be fine to add tests.

sjamboro (commented):
I left some comments. Do we want that error in mocked data? I was able to reproduce this error handling only with the dev environment.

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
vpellegrino: done https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/98/commits/0fc4e8cbb71e3d66db33f5bd738e2981b75f6e40

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
sjamboro: @vpellegrino now I can see that the transformation is editable in `ProcessorDetailPage`. So I confirm that it works here as expected.

PR #99, opened 2022-07-14
[NO-JIRA]: Retrieving the error cause from the first error item

vpellegrino:
As follow up of https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/91 we are going to provide a utility function for retrieving the `reason` field from the first error item and using it in the delete modals.


manstis (approved):
LGTM 👍

PR #100, opened 2022-07-15
MGDOBR-717 - [UI] added test for Instance List pagination

ChughShilpa:
https://issues.redhat.com/browse/MGDOBR-717

### DIFF
--- a/cypress/integration/DeleteTest.ts
+++ b/cypress/integration/DeleteTest.ts
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
 describe("Delete Test", () => {
   it("Instance", () => {
     cy.visit("/");
@@ -29,6 +30,21 @@ describe("Delete Test", () => {
         .eq(2)
         .should("have.text", "deprovision");
 
+=======
+import { deleteInstance } from "./Util";
+
+describe("Delete Test", () => {
+  it("Instance", () => {
+    cy.visit("/");
+    deleteInstance("Instance ten");
+    cy.ouiaId("Instances list table", "PF4/Table").within(() => {
+      // once delete confirmed, state should change
+      cy.ouiaId("Instance ten", "PF4/TableRow")
+        .find("td")
+        .eq(1)
+        .should("have.text", "deprovision");
+
+>>>>>>> 8475636 (Fix review comments)
       // once deprovision is completed, entry should disappear
       cy.ouiaId("Instance ten", "PF4/TableRow", { timeout: 30000 }).should(
         "not.exist"
--- a/cypress/integration/DeleteTest.ts
+++ b/cypress/integration/DeleteTest.ts
@@ -1,36 +1,3 @@
-<<<<<<< HEAD
-describe("Delete Test", () => {
-  it("Instance", () => {
-    cy.visit("/");
-
-    cy.ouiaId("Instances list table", "PF4/Table")
-      .ouiaId("Instance ten", "PF4/TableRow")
-      .find("td")
-      .then(($cells) => {
-        expect($cells).have.length(4);
-        expect($cells.eq(2)).have.text("ready");
-        cy.wrap($cells.eq(3)).ouiaType("PF4/Dropdown").click();
-      });
-
-    cy.ouiaType("PF4/DropdownItem").contains("Delete").click();
-
-    cy.ouiaId("delete-instance", "PF4/ModalContent").within(() => {
-      cy.ouiaId("delete-confirmation-value", "PF4/TextInput").type(
-        "Instance ten"
-      );
-      cy.ouiaId("confirm", "PF4/Button").click();
-    });
-
-    cy.ouiaId("delete-instance", "PF4/ModalContent").should("not.exist");
-
-    cy.ouiaId("Instances list table", "PF4/Table").within(() => {
-      // once delete confirmed, state should change
-      cy.ouiaId("Instance ten", "PF4/TableRow")
-        .find("td")
-        .eq(2)
-        .should("have.text", "deprovision");
-
-=======
 import { deleteInstance } from "./Util";
 
 describe("Delete Test", () => {
@@ -41,10 +8,9 @@ describe("Delete Test", () => {
       // once delete confirmed, state should change
       cy.ouiaId("Instance ten", "PF4/TableRow")
         .find("td")
-        .eq(1)
+        .eq(2)
         .should("have.text", "deprovision");
 
->>>>>>> 8475636 (Fix review comments)
       // once deprovision is completed, entry should disappear
       cy.ouiaId("Instance ten", "PF4/TableRow", { timeout: 30000 }).should(
         "not.exist"
--- a/cypress/integration/InstancesTest.ts
+++ b/cypress/integration/InstancesTest.ts
@@ -1,52 +1,12 @@
 import { format } from "date-fns";
-<<<<<<< HEAD
-=======
 import { createInstance } from "./Util";
 import { deleteInstance } from "./Util";
-<<<<<<< HEAD
-import { newInstanceStatus } from "./Util";
-import { deleteInstanceStatus } from "./Util";
->>>>>>> 8475636 (Fix review comments)
-=======
 import { waitTillInstanceIsReady } from "./Util";
 import { deletedInstanceNotExist } from "./Util";
->>>>>>> 57e9287 (MGDOBR-717 - renaming function name in Util.ts)
 
 const formatDate = (dateStr: string): string =>
   format(new Date(dateStr), "PPPP p");
 
-<<<<<<< HEAD
-function createnewInstance() {
-  const newInstanceName: string = "Some new instance";
-  cy.ouiaId("create-smart-event-instance", "PF4/Button").click();
-  cy.ouiaId("create-instance", "PF4/ModalContent").then(($modal) => {
-    cy.wrap($modal)
-      .should("be.visible")
-      .within(() => {
-        cy.ouiaId("new-name", "PF4/TextInput").type(newInstanceName);
-        cy.ouiaId("info-instance-available-soon", "PF4/Alert").should(
-          "have.text",
-          "Info alert:Your Smart Events instance will be ready for use shortly after creation."
-        );
-        cy.ouiaId("submit", "PF4/Button").click();
-      });
-    cy.wrap($modal, { timeout: 20000 }).should("not.exist");
-  });
-
-  cy.ouiaId("Instances list table", "PF4/Table")
-    .ouiaId(newInstanceName, "PF4/TableRow")
-    .should("be.visible")
-    .within(() => {
-      cy.get("td:first").should("have.text", newInstanceName);
-      cy.get("td:nth-child(2)").then(($state) => {
-        cy.wrap($state).should("have.text", "accepted");
-        cy.wrap($state, { timeout: 30000 }).should("have.text", "ready");
-      });
-    });
-}
-
-=======
->>>>>>> 8475636 (Fix review comments)
 describe("Instances Test", () => {
   describe("the 'Create a SE instance' Modal", () => {
     beforeEach(() => {
@@ -57,54 +17,18 @@ describe("Instances Test", () => {
     });
 
     it("Submit", () => {
-<<<<<<< HEAD
-<<<<<<< HEAD
-<<<<<<< HEAD
       const newInstanceName: string = "Some new instance";
-      cy.ouiaId("create-smart-event-instance", "PF4/Button").click();
-      cy.ouiaId("create-instance", "PF4/ModalContent").then(($modal) => {
-        cy.wrap($modal)
-          .should("be.visible")
-          .within(() => {
-            cy.ouiaId("new-name", "PF4/TextInput").type(newInstanceName);
-            cy.ouiaId("info-instance-available-soon", "PF4/Alert").should(
-              "have.text",
-              "Info alert:Your Smart Events instance will be ready for use shortly after creation."
-            );
-            cy.ouiaId("submit", "PF4/Button").click();
-          });
-        cy.wrap($modal, { timeout: 20000 }).should("not.exist");
-      });
-
+      createInstance(newInstanceName);
       cy.ouiaId("Instances list table", "PF4/Table")
         .ouiaId(newInstanceName, "PF4/TableRow")
         .should("be.visible")
         .within(() => {
           cy.get("td:first").should("have.text", newInstanceName);
           cy.get("td:nth-child(3)").then(($state) => {
-=======
-      createInstance("Some new instance");
-=======
-      const newInstanceName: string = "Some new instance";
-      createInstance(newInstanceName);
->>>>>>> 57e9287 (MGDOBR-717 - renaming function name in Util.ts)
-      cy.ouiaId("Instances list table", "PF4/Table")
-        .ouiaId(newInstanceName, "PF4/TableRow")
-        .should("be.visible")
-        .within(() => {
-          cy.get("td:first").should("have.text", newInstanceName);
-          cy.get("td:nth-child(2)").then(($state) => {
->>>>>>> 8475636 (Fix review comments)
             cy.wrap($state).should("have.text", "accepted");
             cy.wrap($state, { timeout: 30000 }).should("have.text", "ready");
           });
         });
-<<<<<<< HEAD
-=======
-      createnewInstance();
->>>>>>> 429d980 (Replacing existing test logic with a function)
-=======
->>>>>>> 8475636 (Fix review comments)
     });
 
     it("Submit and expect error", () => {
@@ -145,16 +69,7 @@ describe("Instances Test", () => {
         .ouiaId(newInstanceName, "PF4/TableRow")
         .should("not.exist");
     });
-  });
 
-<<<<<<< HEAD
-  describe("the Instance statuses", () => {
-    beforeEach(() => {
-      cy.visit("/");
-      cy.ouiaId("loading-table", "PF4/Card", { timeout: 30000 }).should(
-        "not.exist"
-      );
-=======
     describe("Instance List pagination test", () => {
       it("Select other page", () => {
         cy.get("[aria-label='Go to first page']").should("be.disabled");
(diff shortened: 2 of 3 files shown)


sjamboro (commented):
I am not sure if we could use this approach. It is the first step to create something like Page Fragments which is not recommended by [Cypress - best practice](https://docs.cypress.io/guides/references/best-practices#Organizing-Tests-Logging-In-Controlling-State). The original websites shows plenty of examples how to deal with it. 

I agree that some isolated functions are good to share. IMHO ... delete and create Instance are such functions (two occurences). I think that we can use commands or actions or create separate file (something like Utils.ts). I am not sure if export function from a test file is a good idea. 

Such functions should be:
- parameterized (name, state, etc.). See my comments.
- assert the neccesary objects (does we always assert that the instance is in the appropriate state or wait that the state is changed or number of columns)

Probably large projects are using PageFragments because you can find blog articles like [Stop using Page Objects and Start using App Actions](https://www.cypress.io/blog/2019/01/03/stop-using-page-objects-and-start-using-app-actions/#page-objects-in-cypress).

cypress/integration/InstancesTest.ts
```diff
@@ -79,177 +85,256 @@ describe("Instances Test", () => {
...
 
-  describe("the Instance statuses", () => {
-    beforeEach(() => {
-      cy.visit("/");
-      cy.ouiaId("loading-table", "PF4/Card", { timeout: 30000 }).should(
-        "not.exist"
-      );
-    });
+    it("Instance List pagination test", () => {
+      // cy.ouiaId("OUIA-Generated-Button-plain-3", "PF4/Button").click();       // next button
```
sjamboro, line 90: Do not forget to delete these lines before you merge it

cypress/integration/InstancesTest.ts
```diff
@@ -79,177 +85,256 @@ describe("Instances Test", () => {
...
+      // cy.ouiaId("OUIA-Generated-Button-plain-2", "PF4/Button").click();       // previous button
+      cy.get("[aria-label='Go to first page']").should("be.disabled");
+      cy.get("[aria-label='Go to previous page']").should("be.disabled");
+      cy.get("[aria-label='Go to last page']").should("not.be.disabled");
+      cy.get("[aria-label='Go to next page']").should("not.be.disabled");
+      // Go to next page
+      cy.ouiaId("OUIA-Generated-Button-plain-3", "PF4/Button").click();
```
sjamboro, line 97: Try to avoid using generated ID. It is fine if you can use `[aria-label='.....']`.

cypress/integration/InstancesTest.ts
```diff
@@ -79,177 +85,256 @@ describe("Instances Test", () => {
...
+          expect($cells.eq(1)).have.text("20 per page");
+          expect($cells.eq(2)).have.text("50 per page");
+          expect($cells.eq(3)).have.text("100 per page");
         });
+
+      // page size is changed
+      cy.contains("20 per page").click(); // change page-size
+      cy.get(".pf-c-pagination__total-items >b:nth-of-type(1)").then(function (
+        count
+      ) {
```
sjamboro, line 124: Please unify to use (count => {})

cypress/integration/InstancesTest.ts
```diff
@@ -79,177 +85,256 @@ describe("Instances Test", () => {
...
+        createInstance();
+        cy.get(".pf-c-pagination__total-items >b:nth-of-type(2)").then(
+          function (count) {
+            let instanceCountAfterCreate = parseInt(count.text());
+            expect(instanceCountAfterCreate).to.be.greaterThan(
+              initialInstanceCount
+            );
+          }
+        );
+      });
```
sjamboro, line 148: This peace of code is hard to read.

cypress/integration/InstancesTest.ts
```diff
@@ -1,8 +1,39 @@
+import { countBy } from "cypress/types/lodash";
 import { format } from "date-fns";
+import { deleteInstance } from "./DeleteTest";
```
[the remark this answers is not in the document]
jomarko: it would be better to have some `DeleteUtil` file, or maybe some general `Delete` cusotm cypress command, that would be able to delete anything? instance, source, action ?

jomarko:
I will postpone my review once @sjamboro comments are addressed/responded.

sjamboro:
> I will postpone my review once @sjamboro comments are addressed/responded.

@jomarko can you take a look at [Instance List pagination test](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/deb8d3c05f7cadda1e17c6b82202c8782bf47cfc/cypress/integration/InstancesTest.ts#L89-L164). Does it what you expect?

jomarko (commented):
@sjamboro yes, I think the added test brings reasonable level of checks to the pagination feature. 

However the code looks quite long. So I still wait for your comments to be addressed first to not make the review process 'messy'.

For now I puts just two suggestions how to make the assertions more robust. 

Thank you for the PR @ChughShilpa

cypress/integration/InstancesTest.ts
ChughShilpa: Done

cypress/integration/InstancesTest.ts
ChughShilpa: Done

cypress/integration/DeleteTest.ts
```diff
@@ -1,39 +1,42 @@
-describe("Delete Test", () => {
-  it("Instance", () => {
-    cy.visit("/");
+export function deleteInstance() {
+  cy.ouiaId("Instances list table", "PF4/Table")
+    .ouiaId("Instance ten", "PF4/TableRow")
+    .find("td")
+    .then(($cells) => {
+      expect($cells).have.length(4);
+      expect($cells.eq(1)).have.text("ready");
```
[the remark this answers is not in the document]
ChughShilpa: I couldn't understand this statement. 
As per my understanding, we can delete Instances which are only in "Ready" State

cypress/integration/InstancesTest.ts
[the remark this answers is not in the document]
ChughShilpa: @jomarko @sjamboro 
I created Util.ts file for common functions like CreateInstance, DeleteInstance.
For now, I accounted for Instances only but we can make this more generic like 'Create or Delete anything" in future

cypress/integration/InstancesTest.ts
ChughShilpa: Bifurcated the code into different test cases and i hope it is easy to read now.

cypress/integration/InstancesTest.ts
```diff
@@ -79,177 +85,256 @@ describe("Instances Test", () => {
...
+        count
+      ) {
+        let initialInstanceCount = parseInt(count.text());
+        createInstance();
+        cy.get(".pf-c-pagination__total-items >b:nth-of-type(2)").then(
+          function (count) {
+            let instanceCountAfterCreate = parseInt(count.text());
+            expect(instanceCountAfterCreate).to.be.greaterThan(
+              initialInstanceCount
+            );
```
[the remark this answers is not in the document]
ChughShilpa: > I am not sure if we could use this approach. It is the first step to create something like Page Fragments which is not recommended by [Cypress - best practice](https://docs.cypress.io/guides/references/best-practices#Organizing-Tests-Logging-In-Controlling-State). The original websites shows plenty of examples how to deal with it.
> 
> I agree that some isolated functions are good to share. IMHO ... delete and create Instance are such functions (two occurences). I think that we can use commands or actions or create separate file (something like Utils.ts). I am not sure if export function from a test file is a good idea.
> 
> Such functions should be:
> 
> * parameterized (name, state, etc.). See my comments.
> * assert the neccesary objects (does we always assert that the instance is in the appropriate state or wait that the state is changed or number of columns)
> 
> Probably large projects are using PageFragments because you can find blog articles like [Stop using Page Objects and Start using App Actions](https://www.cypress.io/blog/2019/01/03/stop-using-page-objects-and-start-using-app-actions/#page-objects-in-cypress).



> d

Thanks @jomarko and @sjamboro for the guidance and suggestions :)

Worked on the suggested changes. I preferred creating "Utility file- Util.ts" for common functions ( create and delete instance)
and kept these function as generic one  - a parameterized function with asserting necessary objects only

and hence created different function (newInstanceStatus and deleteInstanceStatus) which basically acts as wait of their expected behaviour (like createInstance should be in ready state and deleteInstance should not exist)

Kindly let me know your opinion on this.

cypress/integration/InstancesTest.ts
ChughShilpa: I tried creating a common method to count Instances to avoid code duplicate but as Cypress is asynchronous and it doesn't work simply with assigning a return value to a variable. 

This is my understanding so far.

cypress/integration/InstancesTest.ts
```diff
@@ -13,27 +17,12 @@ describe("Instances Test", () => {
     });
 
     it("Submit", () => {
-      const newInstanceName: string = "Some new instance";
```
sjamboro, line 16: @ChughShilpa is a problem to leave here this variable?

cypress/integration/Util.ts
```diff
@@ -0,0 +1,55 @@
...
+          "have.text",
+          "Info alert:Your Smart Events instance will be ready for use shortly after creation."
+        );
+        cy.ouiaId("submit", "PF4/Button").click();
+      });
+    cy.wrap($modal, { timeout: 20000 }).should("not.exist");
+  });
+}
+
+export function newInstanceStatus(instanceName: string) {
```
sjamboro, line 18: @ChughShilpa I am a little bit confused by name of this function. When I read code I was not sure about its purpose. I think about `waitInstanceIsReady(instanceName)` or `waitInstanceStatus(instanceName, status)`.
It would be nice to have timeout as optional parameter.

@jomarko what is your opinion?

cypress/integration/Util.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+}
+
+export function deleteInstance(deleteInstanceName: string) {
+  cy.ouiaId("Instances list table", "PF4/Table")
+    .ouiaId(deleteInstanceName, "PF4/TableRow")
+    .find("td")
+    .then(($cells) => {
+      expect($cells).have.length(4);
+      expect($cells.eq(1)).have.text("ready");
```
sjamboro, line 35: @ChughShilpa REST api says that delete is possible only in "ready" or "failed" state. We should allow to define "failed" state here.

cypress/integration/Util.ts
```diff
@@ -0,0 +1,55 @@
...
+  cy.ouiaId("delete-instance", "PF4/ModalContent").within(() => {
+    cy.ouiaId("delete-confirmation-value", "PF4/TextInput").type(
+      deleteInstanceName
+    );
+    cy.ouiaId("confirm", "PF4/Button").click();
+  });
+  cy.ouiaId("delete-instance", "PF4/ModalContent").should("not.exist");
+}
+
+export function deleteInstanceStatus(instanceName: string) {
```
sjamboro, line 49: @ChughShilpa again I am a little bit confused by the name of this function. The name contains "Status" and the code checks that the instance does not exist.

sjamboro (commented):
thanks @ChughShilpa ... I am ready to approved it when you give me answers on my minor comments and when you rebase on the top of the main branch and solve merge conflicts.

You cannot change code but give me a reason why you do it maybe I missed something.

cypress/integration/InstancesTest.ts
ChughShilpa: @sjamboro  As now i created generic function for createInstance and thus passing the instance name as parameter while calling the same function. IMO, there's no need to explicitly assigned it to a constant.

cypress/integration/InstancesTest.ts
```diff
@@ -13,27 +17,12 @@ describe("Instances Test", () => {
...
-            );
-            cy.ouiaId("submit", "PF4/Button").click();
-          });
-        cy.wrap($modal, { timeout: 20000 }).should("not.exist");
-      });
-
+      createInstance("Some new instance");
       cy.ouiaId("Instances list table", "PF4/Table")
-        .ouiaId(newInstanceName, "PF4/TableRow")
+        .ouiaId("Some new instance", "PF4/TableRow")
```
sjamboro, line 22: @ChughShilpa, in case that the `newInstanceName` variable stay in code you cannot change this line.

cypress/integration/InstancesTest.ts
```diff
@@ -13,27 +17,12 @@ describe("Instances Test", () => {
...
         .should("be.visible")
         .within(() => {
-          cy.get("td:first").should("have.text", newInstanceName);
+          cy.get("td:first").should("have.text", "Some new instance");
```
sjamboro, line 25: @ChughShilpa, in case that the newInstanceName variable stay in code you cannot change this line.

cypress/integration/InstancesTest.ts
sjamboro: @ChughShilpa I have created this variable because I needed its value several times and some occurences stays in code even if you created new function. 

Thank you for your answer.

cypress/integration/Util.ts
jomarko: @sjamboro yes, your proposal improves the code readabiloity, +1 to follow your proposal

cypress/integration/InstancesTest.ts
ChughShilpa: i reverted my changes and keep the code as it is

cypress/integration/Util.ts
ChughShilpa: Renamed the function name as suggested

cypress/integration/Util.ts
ChughShilpa: Added "failed" state

cypress/integration/Util.ts
ChughShilpa: modified the function name

jomarko (changes requested):
@ChughShilpa please fix your PR, you committed some merge conflicts strings and your code is not compiling now.

ChughShilpa:
> @ChughShilpa please fix your PR, you committed some merge conflicts strings and your code is not compiling now.



> @ChughShilpa please fix your PR, you committed some merge conflicts strings and your code is not compiling now.

yes am looking into it

ChughShilpa:
> @ChughShilpa please fix your PR, you committed some merge conflicts strings and your code is not compiling now.

Fixed the PR

cypress/integration/InstancesTest.ts
```diff
@@ -79,6 +69,95 @@ describe("Instances Test", () => {
...
+        cy.get("[aria-label='Go to next page']").first().click();
+        cy.get("[aria-label='Go to first page']").should("not.be.disabled");
+        cy.get("[aria-label='Go to previous page']").should("not.be.disabled");
+
+        // as currently count of instances are covered in 2 pages
+        cy.get("[aria-label='Go to last page']").should("be.disabled");
+        cy.get("[aria-label='Go to next page']").should("be.disabled");
+
+        // Go to previous page
+        cy.get("[aria-label='Go to previous page']").first().click();
```
jomarko, line 90: From my point of view this 'previous page' click is not needed, as we do not check the effect of the click action.

- We should either remove it
- Or we should add assertions like we have after clicking 'next page'

cypress/integration/InstancesTest.ts
```diff
@@ -79,6 +69,95 @@ describe("Instances Test", () => {
...
+        // page size is changed
+        cy.contains("20 per page").click();
+        cy.get(".pf-c-pagination__total-items >b:nth-of-type(1)").then(
+          (count) => {
+            let size = count.text();
+            let perPageSize = parseInt(size.slice(-2));
+            cy.get(".pf-c-pagination__total-items >b:nth-of-type(2)").then(
+              (count) => {
+                let pageSize = parseInt(count.text());
+                cy.wrap(perPageSize).should("be.lte", pageSize);
```
jomarko, line 118: Could we please ad assertion that `pageSize` is equal to 20?

jomarko (commented):
Thank you for a PR @ChughShilpa

cypress/integration/InstancesTest.ts
ChughShilpa: I firstly added assertions for current page (first page) as 
`cy.get("[aria-label='Go to first page']").should("be.disabled");
        cy.get("[aria-label='Go to previous page']").should("be.disabled");
        cy.get("[aria-label='Go to last page']").should("not.be.disabled");
        cy.get("[aria-label='Go to next page']").should("not.be.disabled");`
        
Then move to next page and asserted for next page as 
`// Go to next page
        cy.get("[aria-label='Go to next page']").first().click();
        cy.get("[aria-label='Go to first page']").should("not.be.disabled");
        cy.get("[aria-label='Go to previous page']").should("not.be.disabled");

        // as currently count of instances are covered in 2 pages
        cy.get("[aria-label='Go to last page']").should("be.disabled");
        cy.get("[aria-label='Go to next page']").should("be.disabled");`
        
and then move to previous page (which is the same as first page)

jomarko: `and then move to previous page (which is the same as first page) for which we already did the validation`

I agree that previous page is same as first page, where you did some assertions already. However the operation (last line) in your test:
`cy.get("[aria-label='Go to previous page']").first().click();`

How are we sure that after that clic really the fist page is shown? Are we sure there is never ending loading screen? or some other issue?

ChughShilpa: I agree with your on this point. Will add the assertions for the same

cypress/integration/InstancesTest.ts
ChughShilpa: @jomarko 
UI shows the listing of instances as

1. 20 per page
image

2. 10 per page


Thus, pageSize should be equal to 'total no of instances' not pageSize

jomarko: +1

cypress/integration/InstancesTest.ts
```diff
@@ -1,4 +1,8 @@
 import { format } from "date-fns";
+import { createInstance } from "./Util";
+import { deleteInstance } from "./Util";
+import { waitTillInstanceIsReady } from "./Util";
+import { deletedInstanceNotExist } from "./Util";
```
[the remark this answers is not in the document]
ChughShilpa: Done

jomarko (approved):
Thank you for changes

PR #106, opened 2022-07-20
MGDOBR-942: Update delete modal micro copy

vpellegrino:
**Task**: https://issues.redhat.com/browse/MGDOBR-942


src/app/Instance/DeleteInstance/DeleteInstance.tsx
```diff
@@ -54,15 +54,22 @@ const DeleteInstance = (props: DeleteInstanceProps): JSX.Element => {
       setPreloading(false);
       if (processorListResponse.total && processorListResponse.total > 0) {
         setDeleteBlockedReason(
-          t("instance.errors.cantDeleteBecauseProcessorsInside")
+          <Trans
```
manstis, line 57: Out of interest, why do we need `<Trans..>` here where use of `t(..)` works elsewhere?

locales/en/openbridge.json
```diff
@@ -46,9 +46,12 @@
   "instance": {
     "createASEInstance": "Create a Smart Events instance",
     "createSEInstance": "Create Smart Events instance",
+    "delete": "Delete instance",
+    "deleteIt": "Delete Smart Events Instance?",
```
manstis, line 50: You use "...Smart Events **i**nstance..." elsewhere. Do you want to be consistent?

manstis (approved):
A couple of minor comments; but nothing serious.

src/app/Instance/DeleteInstance/DeleteInstance.tsx
vpellegrino: Mainly because in the message we need to translate, there is a `<strong>` tag that we need to render via the `<Trans>` component.

sjamboro (approved):
LGTM

src/app/Instance/DeleteInstance/DeleteInstance.tsx
kelvah, line 57: Not great to store jsx inside a state variable. Can you do this in other ways?

src/app/Instance/InstancePage/InstancePage.css
```diff
@@ -1,3 +1,7 @@
 .instance-page__tabs .pf-c-tabs__item.pf-m-current .pf-c-tabs__item-text {
   color: var(--pf-c-tabs__scroll-button--hover--Color);
 }
+
+.instance-page__actions .pf-c-dropdown__menu {
+  width: 13rem;
+}
```
kelvah, line 7: Nitpicking here... Is `min-width` better here?

kelvah (commented):
I think there is something wrong with the deletion dialog inside the overview page.

It's OK the first time you open it. But if you close it and open it again for a different instance, the initial state is wrong and sometimes it doesn't update correctly after the "processors check" completes.

vpellegrino:
@kelvah, thanks for your review and for catching the issue about the wrong state when reopening the delete dialog.
It was caused by the presence of `instanceName` in the dependency array of the `effect` function (https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/106/commits/84666709a995421c7703eb43a768fa1927035afa#diff-7e6b5a76f5ce5adbe14f1e05637a68fc9677261704890eaa55eb267a6510953aR64)

src/app/Instance/DeleteInstance/DeleteInstance.tsx
vpellegrino: Just a brief overview to justify my choice: the delete blocked reason message can be:
1) a message (already translated) returned by API
2) a message w/ interpolation computed on the UI
3) a message w/o interpolation computed on the UI
One concept was starting to be represented by multiple states and lead to complex management (on the `DeleteModal` side).

BTW, I agree with you that it is not recommended to have a JSX as a state, so the solution I can propose to you is to have a string already translated but with potential formatting to be applied (e.g. `<strong>`) on the `DeleteModal`, so I introduced (https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/106/commits/48ac1dfe0c1167184728ffd97e58bf764242a88e) a library to translate a string, with HTML tags, to a JSX element.

Please let me know your opinion.

kelvah: I see you also did an attempt with the `Trans` component to parse the translated string. 
Any drawbacks using it? It seems to do the job.
it could spare us the addition of another dependency just for this.

vpellegrino: It was working (I checked its source code, and behind the scenes, it was using a parsing library, too), with a warning (repeated on every render) of a missing translation key. 
It is probably designed only for translation purposes, so I stopped using it. 
In our case, we need to parse a string, including HTML tags.

There is an alternative (https://codingshower.com/parse-and-convert-raw-html-string-into-react-elements-or-components/), represented by using `dangerouslySetInnerHTML` feature, but it does not automatically sanitize the input, so I thought that the introduction of a library was the best option.

vpellegrino: Yes, we need to add another library: DOMPurify.
Before adding it, let's decide how we want to proceed.

vpellegrino: We will temporarily use the `Trans` component as agreed in a separate discussion.
Since use-cases are increasingly complex, it is acceptable to review and re-design components, particularly the communication between them. This will be done as part of https://issues.redhat.com/browse/MGDOBR-964

kelvah (approved):
Approving as we agreed to proceed with a refactoring in a different ticket.

Thanks @vpellegrino :+1:

PR #115, opened 2022-08-01
[NO-JIRA]: Skeleton when missing endpoint

vpellegrino:
When no endpoint URL is available (instance creation is still in progress) a skeleton will be shown:
image
In that image the missing endpoint has been forced programmatically, but you can simulate this situation with the real API.

### DIFF
--- a/src/app/Instance/InstanceDetails/InstanceDetails.tsx
+++ b/src/app/Instance/InstanceDetails/InstanceDetails.tsx
@@ -11,6 +11,7 @@ import {
   DrawerHead,
   DrawerPanelBody,
   DrawerPanelContent,
+  Skeleton,
   Stack,
   StackItem,
   Text,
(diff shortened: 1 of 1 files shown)


src/app/Instance/InstanceDetails/InstanceDetails.tsx
```diff
@@ -108,16 +109,20 @@ export const InstanceDetails = ({
...
-                    data-ouia-component-id="instance-details-endpoint"
-                    isBlock
-                    isReadOnly
-                    hoverTip={t("common.copy")}
-                    clickTip={t("common.copied")}
-                    variant={ClipboardCopyVariant.inlineCompact}
-                  >
-                    {instance.endpoint}
-                  </ClipboardCopy>
+                  {instance.endpoint ? (
```
sjamboro, line 112: @vpellegrino can you create a test for this condition? Defined vs undefined endpoint

sjamboro (approved):

kelvah (approved):

src/app/Instance/InstanceDetails/InstanceDetails.tsx
vpellegrino: @sjamboro `InstanceDetails` is a page, not a component, so we don't have any unit test for it.
If you want, you can create a task for creating a new e2e test that will cover this scenario, fine for me.

kelvah: @vpellegrino  Technically we can consider `InstanceDetails` as a component by itself. 
I wouldn't say it's a page. Wdyt?

vpellegrino: We have a flexible definition of what we are considering components and what we are considering pages, but AFAIR we agreed that:
- components are elements with logic to be unit-tested,
- pages are containers of information with no logic

Till now, the `InstanceDetails` has been tested by `InstancesTest.ts` for the case when an instance is in the `READY` state. I am proposing to trace a task for covering the case when the instance is not yet `READY`.

Indeed, currently we have no related `test` file for `InstanceDetails`, as we do for `InstancePage`, and `InstanceListPage`. We should definitely define a clear rule and follow it.
Side note: looking at our codebase, I see that `DeleteInstance` has no tests and probably deserves more tests than `InstanceDetails`. WDYT?

kelvah: I agree the "pages" criteria needs to be more clear.

The idea is that page components are the ones responsible to interact directly with APIs. While "presentational" components inside pages receive data via props and display it in the UI. With this separation, pages should be covered by integration tests while other components with unit tests.

As soon as we started adding features I think a bit of flexibility was necessary. Like for the `DeleteInstance` component you mentioned. It exists because we needed the same API logic in more than one page, otherwise I could have placed its content directly inside `InstancesListPage`. Considering that `DeleteInstance` is dealing with APIs, it makes sense to cover it only with integration tests, while the presentational `DeleteModal` inside it is covered by unit tests.

So it's more about distinguishing between "container" components and "presentational" ones.

Does it make any sense?

PR #116, opened 2022-08-01
MGDOBR-980: [UI] Add source definition test

sjamboro:
https://issues.redhat.com/browse/MGDOBR-980

Create new test suite for the configuration of Processor's source.

PR #101, opened 2022-07-15
MGDOBR-872: Manage masked values inside action/source configuration

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-872

This PR depends on:

- https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/863

I'm adding support for the new format of masked values (empty object, `{}`) inside processors.

To test this you will have to check your browser developer tools console and inspect network requests.

Steps to test it:
- `npm run start:mocked` (until the related PR is merged, you can only use the mocked APIs)
- open a bridge and click on create new processor
- select the sink type
- provide a processor name
- select an action with sensitive fields (like `AWS Lambda`)
- fill up the action configuration parameters
- click on the create processor button
- inspect the payload of the POST request that creates the processor. It will include the sensitive fields values as you provided them
- in the processors list page, wait until the new processor gets `ready` and then click on its name
- as the processor detail page opens, inspect the request to `GET /bridges/{bridgeId}/processors/{processorId}`, its response will report an empty object (`{}`) as value for the sensitive fields
- the sensitive values are displayed as a bunch of asterisks in the properties list
- now click on the edit button
- change an optional param like the transformation template
- notice that the sensitive fields in the form are empty, no value is displayed
- save the processor
- inspect the payload of the PUT request that saves the processor; the secret fields values are empty objects (`{}`), meaning the back-end will not overwrite the existing value provided when creating the processor
- finally, go back to the processors list and wait until the processor becomes `ready`
- click on the processor name to open its detail
- click on edit
- change one of the secret parameters by adding a new value
- save the processor
- inspect the payload of the PUT request that saves the processor; the edited secret parameters will have the new value, meaning the back-end will overwrite the previous one.

PR #119, opened 2022-08-04
MGDOBR-971: Verify new Azure Event Hubs source schema

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-971

### DIFF
--- a/mocked-api/schemasData.ts
+++ b/mocked-api/schemasData.ts
@@ -87,6 +87,14 @@ export const schemaCatalogData = [
     type: "source",
     href: "/api/smartevents_mgmt/v1/schemas/sources/google_pubsub_source_0.1",
   },
+  {
+    kind: "ProcessorSchemaEntry",
+    id: "azure_eventhubs_source_0.1",
+    name: "Azure EventHub Source",
+    description: "Ingest data from Azure EventHub.",
+    type: "source",
+    href: "/api/smartevents_mgmt/v1/schemas/sources/azure_eventhubs_source_0.1",
+  },
 ];
 
 export const schemasData: { [key: string]: object } = {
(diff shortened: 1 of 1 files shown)


vpellegrino (approved):
LGTM

manstis (approved):
LGTM :+1:

mocked-api/schemasData.ts
```diff
@@ -785,4 +793,80 @@ export const schemasData: { [key: string]: object } = {
...
+      azure_eventhub_name: {
+        title: "Eventhubs Name",
+        description: "The Event Hub name.",
+        type: "string",
+      },
+      azure_shared_access_name: {
+        title: "Share Access Name",
+        description: "The Event Hubs SAS key name.",
+        type: "string",
+        "x-group": "credentials",
```
sjamboro, line 823: @kelvah please can you share a definition about "x-group:gredentials" I thought that it marks that the parameter should be hidden but the value is still visible. See: 
image

mocked-api/schemasData.ts
```diff
@@ -785,4 +793,80 @@ export const schemasData: { [key: string]: object } = {
...
+      },
+      azure_shared_access_key: {
+        title: "Share Access Key",
+        "x-group": "credentials",
+        oneOf: [
+          {
+            title: "Share Access Key",
+            description: "The key for the EventHubs SAS key name.",
+            type: "string",
+            format: "password",
```
sjamboro, line 833: @kelvah probably this is not a problem of this issue but the value of this property is visible in Process Details. Is it a know issue?
image

sjamboro (approved):
thanks @kelvah ... I have left minor comments.

sjamboro:
@kelvah there will be problem with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/116. Do you want to merge this first?

mocked-api/schemasData.ts
kelvah: What makes a parameter hidden is its actual definition. 
It has to include the `oneOf` property with the double field definition: one with `format="password"` and the other with `type="object"`
You can see an example here:
https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/8d96ac75b463442d90c9cd277935847271d1d3e5/mocked-api/schemasData.ts#L205-L221

I'm not sure what the `x-group` property stands for. It's clearly not a json-schema prop. We inherit these from the original kamelet used to implement the connector. See [here for example](https://github.com/apache/camel-kamelets/blob/main/kamelets/azure-eventhubs-source.kamelet.yaml#L53-L67). But we are not actually using them.

mocked-api/schemasData.ts
kelvah: This will be covered by #101.

kelvah:
> @kelvah there will be problem with #116. Do you want to merge this first?

OK. Thanks.

PR #123, opened 2022-08-11
MGDOBR-984: Add Error Handling configuration to the bridge creation dialog

vpellegrino:
**Task**: https://issues.redhat.com/browse/MGDOBR-984
image


src/app/Instance/CreateInstance/ErrorHandlingSelection.tsx
```diff
@@ -0,0 +1,76 @@
...
+        {errorHandling.label}
+      </SelectOption>
+    ));
+  }, [errorHandlingMethods.deadLetterQueue]);
+
+  return (
+    <Select
+      id="method-selector"
+      isDisabled={isDisabled}
+      ouiaId="method-selector"
```
jomarko, line 52: Could we please update `ouiaId` to provide more information? `method-selector` sounds to me too general. Or do we wrap it into some parent component that gives more context to this `ouiaId`. I like `fieldId={"error-handling-method"}` from `CreateInstance.tsx`

jomarko (commented):
One comment after code review. Manual review pending.

jomarko (commented):
Question coming from manual review. 

In the design document, we show how to view details of 'Error Handling strategy' for given smart event instance. However I am not able to see such details once I start app as `npm run start`. 

Am I correct we split the tasks and it is expected? Do we track this in some jira?

src/app/Instance/CreateInstance/ErrorHandlingSelection.tsx
vpellegrino: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/123/commits/0628f8da8f2eed9e00e852bde4a9dee7d4b4fea7

vpellegrino:
> Question coming from manual review.
> 
> In the design document, we show how to view details of 'Error Handling strategy' for given smart event instance. However I am not able to see such details once I start app as `npm run start`.
> 
> Am I correct we split the tasks and it is expected? Do we track this in some jira?

In the epic https://issues.redhat.com/browse/MGDOBR-893 we already have all tasks related to error handling.
For example, https://issues.redhat.com/browse/MGDOBR-985 is about display the error handling strategy for a given smart events instance.

vpellegrino:
@jomarko there is an e2e test failing, can you please help me to understand the reason? The logs do not provide info + I am not able to reproduce it locally

vpellegrino:
I checked the same job on https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/127 and we have the same error (https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/runs/7875916762?check_suite_focus=true) as there (https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/runs/7874041645?check_suite_focus=true), so we can state that the issue is not related to this change.

sjamboro:
@vpellegrino there is a screenshot of failing test. My opinion is that the problem is in the environment. The error message says "Timed out retrying after 60000ms: expected '<td>' to have text 'ready', but the text was 'failed' ".
It seems that this GHA was failing from Tue to Wed. See https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/actions/workflows/dev-cluster-long-running-bridge.yml
 I will rerun E2E tests.
image

vpellegrino:
@kelvah @manstis @jomarko I just solved conflicts with `main`. Could you please review this PR?

src/app/Instance/CreateInstance/CreateInstance.tsx
```diff
@@ -309,6 +391,37 @@ const CreateInstance = (props: CreateInstanceProps): JSX.Element => {
...
+                setErrorHandlingParameters({});
+                if (errorMethod === ERROR_HANDLING_METHODS.default.value) {
+                  setErrorHandlingSchemaId(null);
+                } else {
+                  setErrorHandlingSchemaId(errorMethod);
+                }
+              }}
+            />
+          </FormGroup>
+          {errorHandlingSchema && !errorHandlingSchemaLoading && (
```
manstis, line 414: This is causing the form to flicker when selecting between types.

IDK whether it's better to just check if the `setErrorHandlingSchemaId != null`?

manstis (approved):
A minor cosmetic comment.

Personally, showing "Dead Letter Queue" is unnecessary as RHOSE uses `dlq` internally when configuring Managed Connectors (vs `ignore` or `log`). It is not something visible to Users nor something they can change. That is controlled by the selector ("Ignore", "WebHook" or "Kafka Topic")... but I see the design document shows it; so...

kelvah (commented):
Thanks @vpellegrino, nice work.

I've noticed that when submitting the form with the bridge name empty and an EH method selected, the validation reports an error on the name field only. The errors in the EH section are displayed only after fixing the name.
The same behavior happens when there's the "bridge name already used" error.

manstis:
> Thanks @vpellegrino, nice work.
> 
> I've noticed that when submitting the form with the bridge name empty and an EH method selected, the validation reports an error on the name field only. The errors in the EH section are displayed only after fixing the name. The same behavior happens when there's the "bridge name already used" error.

This is a manifestation of the RHOSE-wide validation issue (where we don't validate everything at once).

See https://issues.redhat.com/browse/MGDOBR-947

src/app/Instance/CreateInstance/CreateInstance.tsx
vpellegrino: Hi @manstis, I tried to increase the delay for the schema API response, and it seems that the "flicking" effect is because when changing the method, the section disappears. As soon as we get the API response, the section appears again (with the proper content).
changing_method_delayed

We have the same condition as in the processor edit page. I am unsure if we should handle it differently here since we have a modal with a dynamic height length.

manstis:
> > Thanks @vpellegrino, nice work.
> > I've noticed that when submitting the form with the bridge name empty and an EH method selected, the validation reports an error on the name field only. The errors in the EH section are displayed only after fixing the name. The same behavior happens when there's the "bridge name already used" error.
> 
> This is a manifestation of the RHOSE-wide validation issue (where we don't validate everything at once).
> 
> See https://issues.redhat.com/browse/MGDOBR-947

Actually, I am writing a quick unit test for the backend this does not appear to be the case.

I am seeing two errors returned from the backend (missing name and missing Error Handler parameters).

Please see my backend PR: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1066

vpellegrino:
> A minor cosmetic comment.
> 
> Personally, showing "Dead Letter Queue" is unnecessary as RHOSE uses `dlq` internally when configuring Managed Connectors (vs `ignore` or `log`). It is not something visible to Users nor something they can change. That is controlled by the selector ("Ignore", "WebHook" or "Kafka Topic")... but I see the design document shows it; so...

You definitely have a point. In this case, I followed the design doc. I think we can change it very easily.

src/app/Instance/CreateInstance/CreateInstance.tsx
manstis: Fair enough. It is cosmetic and, if we have the same behaviour elsewhere, I will ignore (I have approved this PR!!)

When the selection changes we end up setting `errorHandlingSchemaLoading` to `true` meaning the form section is not rendered until the new schema retrieved and parsed. If we only checked `setErrorHandlingSchemaId != null` we'd show the existing schema form while loading the new one meaning no flicker. You could add `errorHandlingSchema && !errorHandlingSchemaLoading` to the `isDisabled` check resulting in the existing schema form being disabled while the new one loads to prevent naughty users from doing anything. Just a thought. No need to follow-up. PR approved "as is".

kelvah:
> > A minor cosmetic comment.
> > Personally, showing "Dead Letter Queue" is unnecessary as RHOSE uses `dlq` internally when configuring Managed Connectors (vs `ignore` or `log`). It is not something visible to Users nor something they can change. That is controlled by the selector ("Ignore", "WebHook" or "Kafka Topic")... but I see the design document shows it; so...
> 
> You definitely have a point. In this case, I followed the design doc. I think we can change it very easily.

I guess the "DLQ" label was intentionally added to make our EH comparable to the one happening within COS, where there are other options... I guess we can re-discuss it in the ux sync.

vpellegrino:
> Thanks @vpellegrino, nice work.
> 
> I've noticed that when submitting the form with the bridge name empty and an EH method selected, the validation reports an error on the name field only. The errors in the EH section are displayed only after fixing the name. The same behavior happens when there's the "bridge name already used" error.

I think I've found the cause. We always need to execute the  `validateParameters.current?.()` function to report the errors in the EH section.
I just fixed it. Please, have a look when it's possible.

jomarko (commented):
I didn't found any issue during review. However I would like to check if error handling conf we introduced as part of https://issues.redhat.com/browse/MGDOBR-984 is compatible with future plans https://issues.redhat.com/browse/MGDOBR-1009. 

In other words, should I prepare some cypress tests that creates bridge with real error handler configured or the mechanism will be changed completely soon? Will MGDOBR-1009 just extend current mechanism and it is safe to test?

vpellegrino:
> I didn't found any issue during review. However I would like to check if error handling conf we introduced as part of https://issues.redhat.com/browse/MGDOBR-984 is compatible with future plans https://issues.redhat.com/browse/MGDOBR-1009.
> 
> In other words, should I prepare some cypress tests that creates bridge with real error handler configured or the mechanism will be changed completely soon? Will MGDOBR-1009 just extend current mechanism and it is safe to test?

@jomarko unfortunately I am not able to answer this question. 
In the meantime, we can postpone the cypress tests creation, open a dedicated JIRA and work on it as soon as the situation becomes more clear. WDYT?

kelvah:
> I think I've found the cause. We always need to execute the `validateParameters.current?.()` function to report the errors in the EH section. I just fixed it. Please, have a look when it's possible.

Thanks for fixing this!

I noticed another thing :eyes: 

Name already existing + correct EH configuration scenario. After submitting the form there is a quick loading and then nothing happens. After submitting again the dialog scrolls to the top where the "existing name" error is visible.

vpellegrino:
Nice catch @kelvah 
The issue was related to the missing DOM node (`.pf-m-error`) when the API response generates the error.
Indeed, the element `.pf-m-error` is added after `nameError` is updated (https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/123/commits/d71a30645024fc86eb0fefdf3dc516663f54988e)

jomarko (approved):
Thank you @vpellegrino . I created task https://issues.redhat.com/browse/MGDOBR-1038

kelvah:
> Nice catch @kelvah The issue was related to the missing DOM node (`.pf-m-error`) when the API response generates the error. Indeed, the element `.pf-m-error` is added after `nameError` is updated ([d71a306](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/commit/d71a30645024fc86eb0fefdf3dc516663f54988e))

@vpellegrino :+1: 
I'm wondering if you have to do the same for the generic error too. It's probably the same scenario of the `nameError`.

vpellegrino:
> > Nice catch @kelvah The issue was related to the missing DOM node (`.pf-m-error`) when the API response generates the error. Indeed, the element `.pf-m-error` is added after `nameError` is updated ([d71a306](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/commit/d71a30645024fc86eb0fefdf3dc516663f54988e))
> 
> @vpellegrino +1 I'm wondering if you have to do the same for the generic error too. It's probably the same scenario of the `nameError`.

yes, because even `genericError` is set after API response and not on forum submit. https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/123/commits/36f8df2ec2121a9a856e12ef0f13072225e750c4

kelvah:
Problem is gone with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/123/commits/dce17ee1148f423f536c17f62cc08c1d524f6406.

kelvah (approved):

PR #127, opened 2022-08-16
MGDOBR-865: Run same test with mocked data and real data

sjamboro:
https://issues.redhat.com/browse/MGDOBR-865

This is a draft PR.
Currently there should be only minor changes which leads to unification of test suittes (mocked / e2e).

**Goals:**
- successful run of: `Smart Events UI :: E2E :: Quality Checks` (renamed to `... UI :: DEV :: Quality ...`)
- successful run of: `Smart Events UI :: Quality Checks` (renamed to `... UI :: Mocked :: Quality ...`)
- run similar tests inside 
- possible to define that test run only on one environment

**Current state:**
All tests are integrated together. Compare:
**_Mocked env_**
- REST isskiped
- BasicTest contains 4 tests
- InstanceTest contains 14 tests

image

**_Dev env_**
- Action, Source, Delete, Processor Test are skipped
- BasicTest contains just 3 tests
- InstanceTest contains 3 tests

image


cypress/cypress.e2e.json
```diff
@@ -16,6 +16,7 @@
   "viewportHeight": 1080,
   "env": {
     "REST_URL": "https://event-bridge-event-bridge-prod.apps.openbridge-dev.fdvn.p1.openshiftapps.com",
-    "REST_PATH": "/api/smartevents_mgmt/v1"
+    "REST_PATH": "/api/smartevents_mgmt/v1",
+    "TEST_TYPE": "dev"
```
sjamboro, line 20: @jomarko @ChughShilpa,
Config files are different only in few options as `"env": {"TEST_TYPE": "mocked"}` and `"baseUrl": "http://localhost:8080" }`
The rest of options is same. Do you prefer using of two separate config files or format like this https://docs.cypress.io/guides/references/configuration#Testing-Type-Specific-Overrides? Can I upgrage cypress in this JIRA?

sjamboro: FYI - The migration guide: https://docs.cypress.io/guides/references/migration-guide

I found out that it reflects new changes in JavaScript.

sjamboro: FYI - The old post on Cypress blog (timestamp 2 years ago - about v[4.9.x](https://docs.cypress.io/guides/references/changelog#4-9-0)): https://www.cypress.io/blog/2020/06/18/extending-the-cypress-config-file/
```Cypress does not support extends syntax in its [configuration file](https://on.cypress.io/configuration). If you want to apply different settings, you need to write a complete second configuration file and use it via the --config-file <filename> command line argument.```

File: cypress/integration/Util.ts [hunk 2]
```diff
@@ -57,3 +57,35 @@ export function deletedInstanceNotExist(instanceName: string) {
     );
   });
 }
+
+/**
+ * Assert if the test environment is set by the expected value.
+ * @param type a name of the desired test environment
+ * @returns
+ */
+export function isEnvironmentType(type: string) {
```
jomarko, line 66: can't we use `type: EnvType` instead of `type: string`?

jomarko (commented):
thank you @sjamboro 

I like smaller PRs with single goal/responsibility. As this time, it seems cypress upgrade will require also code changes, I prefer to do it separately.

Regarding two config files vs test specific config. I do not have enough knowledge into this maybe, but two config files seems to me more suitable. It seems to me more aligned with the way of running the tests:
```
    "cypress:run:it": "cypress run --browser chrome --project cypress --config-file cypress.it.json",
    "cypress:open:e2e": "cypress open --browser chrome --project cypress --config-file cypress.e2e.json",
```

ChughShilpa (commented):
Thank you @sjamboro for PR

I also prefer to do cypress upgrade in a separate PR as this will require code changes.

and regarding two config files vs single 'Testing Type-Specific' config file, two different config files seem to be more convenient otherwise every time while running test for different environment we need to change the environment variable in CLI

ChughShilpa (approved):

cypress/integration/Util.ts
sjamboro: fixed

cypress/tsconfig.json
```diff
@@ -8,6 +8,8 @@
     "support/**/*.ts",
     "integration/**/*.ts",
     "e2e/**/*.ts",
+    "mocked/**/*.ts",
```
sjamboro, line 11: This should be deleted

cypress/utils/Util.ts
```diff
@@ -0,0 +1,114 @@
...
+    return `testui-${prefix}-${Cypress._.uniqueId(Date.now().toString())}`;
+  } else {
+    throw new Error(
+      "The environment type is not recognized. Values of CYPRESS_TEST_TYPE are defined by the 'EnvType' enum."
+    );
+  }
+}
+
+export function pageWasLoaded() {
+  cy.get("#nav-toggle", { timeout: 30000 }).should("be.visible");
```
sjamboro, line 14: I have added wait on some basic element to be sure that page is loaded correctly.
Otherwise next waiting rutine (loading-table) was succesful even if the page was empty.

cypress/utils/Util.ts
```diff
@@ -0,0 +1,114 @@
...
+      );
+      cy.ouiaId("new-name", "PF4/TextInput").type(newInstanceName);
+      cy.ouiaId("info-instance-available-soon", "PF4/Alert").should(
+        "have.text",
+        "Info alert:Your Smart Events instance will be ready for use shortly after creation."
+      );
+      cy.ouiaId(action ?? "submit", "PF4/Button").click();
+    });
+    cy.wrap($modal, { timeout: 20000 }).should("not.exist");
+  });
```
sjamboro, line 38: I found out that we can use aliases instead of `then`. Here is the example of such code:
```
  cy.ouiaId("create-smart-event-instance", "PF4/Button").click();
  cy.ouiaId("create-instance", "PF4/ModalContent").as("modal");
  cy.get("@modal")
    .should("be.visible")
    .ouiaId("cloud-provider", "skeleton")
    .should("not.exist");
  cy.get("@modal")
    .ouiaId("cloud-region","PF4/Select", {timeout: 20000}).should("not.have.text", "Select Region");
  cy.get("@modal").within(() => {
    cy.ouiaId("new-name", "PF4/TextInput").as("new-name");
    cy.get("@new-name").type(newInstanceName);
    cy.ouiaId("info-instance-available-soon", "PF4/Alert").as("info");
    cy.get("@info").should(
      "have.text",
      "Info alert:Your Smart Events instance will be ready for use shortly after creation."
    );
    cy.ouiaId(action ?? "submit", "PF4/Button").click();
  });
  cy.get("@modal", { timeout: 20000 }).should("not.exist");
```

sjamboro, line 38: I found out that we can use aliases instead of `then`. Here is the example of such code:
```
  cy.ouiaId("create-smart-event-instance", "PF4/Button").click();
  cy.ouiaId("create-instance", "PF4/ModalContent").as("modal");
  cy.get("@modal")
    .should("be.visible")
    .ouiaId("cloud-provider", "skeleton")
    .should("not.exist");
  cy.get("@modal")
    .ouiaId("cloud-region","PF4/Select", {timeout: 20000}).should("not.have.text", "Select Region");
  cy.get("@modal").within(() => {
    //TODO: inspect why the DOM is detached here
    cy.ouiaId("new-name", "PF4/TextInput").as("new-name");
    cy.get("@new-name").type(newInstanceName);
    cy.ouiaId("info-instance-available-soon", "PF4/Alert").as("info");
    cy.get("@info").should(
      "have.text",
      "Info alert:Your Smart Events instance will be ready for use shortly after creation."
    );
    cy.ouiaId(action ?? "submit", "PF4/Button").click();
  });
  cy.get("@modal", { timeout: 20000 }).should("not.exist");
```

cypress/utils/Util.ts
```diff
@@ -0,0 +1,114 @@
...
+        "Select Region"
+      );
+      cy.ouiaId("new-name", "PF4/TextInput").type(newInstanceName);
+      cy.ouiaId("info-instance-available-soon", "PF4/Alert").should(
+        "have.text",
+        "Info alert:Your Smart Events instance will be ready for use shortly after creation."
+      );
+      cy.ouiaId(action ?? "submit", "PF4/Button").click();
+    });
+    cy.wrap($modal, { timeout: 20000 }).should("not.exist");
```
sjamboro, line 37: Someting

vpellegrino:
@sjamboro It is not clear to me if review is really READY to be reviewed or not.
Please ping me when all pending comments are solved.

jomarko (commented):
Thank you for a PR @sjamboro . I put some comments inline. Also just FYI we need to refactor clean up hook, see more details in https://issues.redhat.com/browse/MGDOBR-1034

cypress/utils/Util.ts
sjamboro: @sjamboro I have no strong opinion. I am not familiar with life-cycle of aliases. It could bring more problems than benefits. I [found out](https://www.youtube.com/watch?v=VPznmFpa1Jc) that it could be useful to avoid the `detached from DOM` error. So I rewrite it and find out that code is better readable for me. But aliases were not useful (test fails again but not so offten) then I have defined `pageWasLoaded` and it fixed all occurence of the `detached from DOM` error. So let leave decision on a creator of a test.

cypress/integration/smoke/RESTInteractionTest.ts
```diff
@@ -0,0 +1,117 @@
...
+          region: "us-east-1",
+        },
+      };
+      cy.request(options).then((response) => {
+        // response.body is automatically serialized into JSON
+        expect(response.body).to.not.be.null;
+        expect(response.body.id).to.not.be.null;
+        bridgeId = response.body.id;
+      });
+    });
```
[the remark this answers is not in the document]
sjamboro: @jomarko I hope that I undrestand the question correctly.
Current state:
- no pre-condition -> run on both environment (Mocked and DEV)
- pre-condition -> define which environment is a target (Mocked xor DEV)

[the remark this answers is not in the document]
jomarko: @sjamboro yes, I understand it as you describe.

However I think `BasicTest` is running on both mocked and DEV, so I am not sure why we do not need there this "TOKEN and variables initialization part" that we have in `RESTInteractionTest`

cypress/utils/Util.ts
sjamboro: @jomarko I have try to put some description here: https://github.com/sjamboro/sandbox-ui/blob/MGDOBR-865/cypress/README.md#how-to-create-tests

Should I rewrite it or highligths important information?

cypress/integration/smoke/RESTInteractionTest.ts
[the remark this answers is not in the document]
sjamboro: OK ... I will check it. `token` is only for REST. I should check `user` and `password`.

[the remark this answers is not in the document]
sjamboro: I do not think that we need `if (typeof user !== "string" || !user) {..}` check for all test but once it is helpfull. Because it read ENV and I suppose that it is stable for all executions. Maybe move this checks under a new test in Basic test running only with DEV. The name should be `Check environment`.

cypress/utils/Util.ts
jomarko: @sjamboro I think it is fine to have a description/guide on place you pointed to, resolving this conversation, thank you

cypress/integration/smoke/RESTInteractionTest.ts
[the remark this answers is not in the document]
jomarko: I think what you mention is OK for CI where we run all tests, however for local development and running just single tets usually, I am not sure if user will be informed what variables were not set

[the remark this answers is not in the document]
sjamboro: @jomarko I have posponed it to https://issues.redhat.com/browse/MGDOBR-1041

vpellegrino (approved):
Generally speaking I link this approach. I was just wondering if it is possible to get rid of e2e/IT, in place of dev/mocked.
I am expecting consistency also in file names.
WDYT?

cypress/integration/smoke/RESTInteractionTest.ts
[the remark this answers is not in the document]
jomarko: thank you @sjamboro

sjamboro:
@vpellegrino good question we should talk about it. During proccess of writing our set of IT and E2E tests we ussually ask if this test can run with real data. Of course the conversion of a test from IT to a E2E test is not 1 to 1. I guess that 80 % of code is same (why duplicate it). Definitely, we need to solve test data set and clear data on the dev stage.

I think that IT test suite running with Mocked environment will be larger than E2E test suite running with the dev stage. IMHO there should be rule that one file contains all three groups (both, only mocked and only dev) of tests. I expect that file names will be consistent.

As a writer of these tests the marking of Dev and Mocked environment in each test helps me to recall all differences. First I start using IT and E2E but it was painfull to me. Because UI is same in both cases only difference is now the environment.

I have no problem to leave IT / E2E to be consistent with other repositories. I have tried to use it different.

vpellegrino:
> @vpellegrino good question we should talk about it. During proccess of writing our set of IT and E2E tests we ussually ask if this test can run with real data. Of course the conversion of a test from IT to a E2E test is not 1 to 1. I guess that 80 % of code is same (why duplicate it). Definitely, we need to solve test data set and clear data on the dev stage.
> 
> I think that IT test suite running with Mocked environment will be larger than E2E test suite running with the dev stage. IMHO there should be rule that one file contains all three groups (both, only mocked and only dev) of tests. I expect that file names will be consistent.
> 
> As a writer of these tests the marking of Dev and Mocked environment in each test helps me to recall all differences. First I start using IT and E2E but it was painfull to me. Because UI is same in both cases only difference is now the environment.
> 
> I have no problem to leave IT / E2E to be consistent with other repositories. I have tried to use it different.

Ok I got the difference.
Fair enough.
Thank you

cypress/utils/Util.ts
ChughShilpa: I found code is readable in both the cases (then and as ). To me , it is hardly making any difference from reading point of view.
But after reading links shared by Jozef, i agree with him that as seems not be much beneficial

sjamboro:
@jomarko @ChughShilpa can you review this PR? Did I resolve your comments?

jomarko (approved):

cypress/integration/smoke/BasicTest.ts
```diff
@@ -1,7 +1,10 @@
+import { onlyOn } from "@cypress/skip-test";
+import { EnvType, isEnvironmentType, safeLogin } from "../../utils/Util";
+
 describe("Basic Elements", () => {
   beforeEach(() => {
     cy.visit("/");
-
+    safeLogin();
     cy.ouiaId("loading-table", "PF4/Card", { timeout: 30000 }).should(
```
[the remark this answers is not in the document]
sjamboro: Good catch ... thanks

PR #146, opened 2022-08-26
Update Renovate according to kas-ui, srs-ui and kafka-ui

jomarko:
- Ui modules of managed services https://github.com/bf2fc6cc711aee1a0c2a usess slightly different renovate configuration as sandbox-ui is using currently.
- Renovate community advice to use property for approving upgrades in dashboard https://github.com/renovatebot/renovate/discussions/17426#discussioncomment-3482988

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,5 @@
 {
   "extends": ["github>redhat-developer/app-services-renovate-config"],
-  "dependencyDashboardApproval": true,
   "separateMajorMinor": true,
   "separateMultipleMajor": true,
   "packageRules": [
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,8 @@
 {
   "extends": ["github>redhat-developer/app-services-renovate-config"],
+  "dependencyDashboardApproval": true,
+  "separateMajorMinor": true,
+  "separateMultipleMajor": true,
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": ["github>redhat-developer/app-services-renovate-config"],
+  "dependencyDashboardApproval": true,
```
kelvah, line 3: Setting this as `true` will prevent renovate from creating any PR automatically.
Maybe you want to set it only for major updates?

Anyway, I think it's a matter of deciding how to deal with majors. We could keep the current config and leave PRs like the Jest one open until they become ready to be merged (when the related packages are updated too). 
Otherwise, we could decide to manually trigger the PR creation for every major release, like you propose.

Maybe the second approach is more work? Meaning monitoring periodically the dashboard and manually trigger PRs

renovate.json
```diff
@@ -1,5 +1,8 @@
...
+  "separateMajorMinor": true,
+  "separateMultipleMajor": true,
```
kelvah, line 5: +1 on these two. I see `separateMajorMinor` should be `true` by default but perfectly fine to make it explicit

jomarko:
@kelvah done

kelvah (approved):
LGTM

PR #147, opened 2022-08-26
Remove 'monaco-editor' dependency

jomarko:
'npx depcheck' marked 'monaco-editor' as unused dependnecy

My guess we use 'monaco-editor' just for processor template? It seems still up and running.
Screenshot from 2022-08-26 15-52-52

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,6 @@
         "lodash.isequal": "4.5.0",
         "lodash.isequalwith": "4.4.0",
         "lodash.omit": "4.5.0",
-        "monaco-editor": "0.21.3",
         "monaco-editor-webpack-plugin": "2.1.0",
         "react": "17.0.2",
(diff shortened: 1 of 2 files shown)


kelvah:
You're right, it seems unnecessary.

It is not listed in the [peer dependencies list](https://github.com/patternfly/patternfly-react/blob/862a38b7f3fccd6f3bfb3ba754d02f03a152dc04/packages/react-code-editor/package.json#L39) of our version of @patternfly/react-code-editor

I've probably added it because it is **wrongfully** reported as [peer dep in the readme](https://github.com/patternfly/patternfly-react/tree/862a38b7f3fccd6f3bfb3ba754d02f03a152dc04/packages/react-code-editor#usage)

kelvah (approved):

vpellegrino (approved):
+1 for removing it, since it doesn't seem to be used

PR #171, opened 2022-09-02
[NO-JIRA] Fix TableRow type inside RenderActions

kelvah:
Removing ts generics usage within `RenderActions` because it's not really necessary and it's causing issues with renovate while updating typescript.

### DIFF
--- a/src/app/components/Table/Table.tsx
+++ b/src/app/components/Table/Table.tsx
@@ -26,11 +26,11 @@ export interface TableRow extends IRowData {
   originalData?: IRowData;
 }
 
-export type RenderActions = <TRow>(props: {
+export type RenderActions = (props: {
   /** PF ActionsColumn component used to render the actions  */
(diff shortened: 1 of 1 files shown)


src/app/components/Table/Table.tsx
```diff
@@ -26,11 +26,11 @@ export interface TableRow extends IRowData {
   originalData?: IRowData;
 }
 
-export type RenderActions = <TRow>(props: {
+export type RenderActions = (props: {
   /** PF ActionsColumn component used to render the actions  */
   ActionsColumn: typeof ActionsColumn;
   /** The row data passed to the actions */
-  row: TRow;
+  row: TableRow;
```
kelvah, line 33: We don't need a generic type here as we will always deal with `TableRow` data.

vpellegrino (approved):
Thank you @kelvah

jomarko (approved):

PR #173, opened 2022-09-05
MGDOBR-1059: Optional secrets value displayed as masked value even when the value was not set

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1059

Steps to test:
- Create a bridge with webhook error handler **without** providing a user/password combination
- Wait for the bridge to get ready
- Open the bridge, click on the `Error handling` link
- The password field value is displayed as `Property not configured` (it was a masked value before)

Same thing can be tested on a sink processor with webhook action. It's the same component used there too.


jomarko (approved):
Thank you @kelvah . This is a nice improvement.

vpellegrino (approved):
LGTM

PR #186, opened 2022-09-20
NO-JIRA: Add cloud providers info to mocked bridges responses

kelvah:
Adding missing `cloud_provider` and `region` properties to bridges mocked API responses.

Also realized `owner` property was not set for new bridges. Fixed it too.


jomarko (approved):
Fine for me. Thank you for an update.

PR #199, opened 2022-09-28
MGDOBR-1108: Red Hat font not loading properly

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1108

Issue was caused by the update to `css-loader@6.7.1` introduced with #164.

As reported by `css-loader` [release notes](https://github.com/webpack-contrib/css-loader/releases/tag/v6.0.0), some changes to the webpack configuration for assets fix the problem.

To test that the Red Hat font is properly used, inspect a text node with chrome developer tools and look for the computed tab.

___

Before

before  

After

after

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -51,39 +51,27 @@ module.exports = (env, argv) => {
         },
         {
           test: /\.(ttf|eot|woff|woff2)$/,
-          use: {
-            loader: "file-loader",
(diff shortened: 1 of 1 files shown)


vpellegrino (approved):

jomarko (approved):
Well spotted. Approving.

PR #203, opened 2022-09-29
MGDOBR-1095: Add error handling strategy switch tests

jomarko:
For more details see: https://issues.redhat.com/browse/MGDOBR-1095

This PR is in a 'draft' due to some internal refactoring done in #193 and fix in #207

### DIFF
--- a/cypress/integration/instance/InstanceEdit.ts
+++ b/cypress/integration/instance/InstanceEdit.ts
@@ -88,25 +88,10 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
         .within(() => {
           cy.get("td:first").should("contain.text", "Instance one");
           //The status column is loading and contains ready
-          cy.get("td:nth-child(3)")
-            .ouiaId("accepted", "QE/ResourceStatus")
-            .within(() => {
-              cy.get("span[role='progressbar']").should("exist");
-              cy.ouiaId("creating", "PF4/Button").should(
-                "have.text",
-                "Creating"
-              );
-              cy.ouiaId("ready-shortly", "QE/HelperTextItem").should(
-                "have.text",
-                "This will be ready shortly."
-              );
-            });
-          cy.ouiaId("creating", "PF4/Button").click();
+          cy.get("td:nth-child(3)").ouiaId("ready", "QE/ResourceStatus", {
+            timeout: 90000,
+          });
         });
-
-      cy.ouiaId("se-status", "QE/Popover", { timeout: 30000 }).should(
-        "not.exist"
-      );
     });
   });
 });
--- a/cypress/integration/instance/InstanceEdit.ts
+++ b/cypress/integration/instance/InstanceEdit.ts
@@ -4,20 +4,12 @@ import { isEnvironmentType, EnvType, pageWasLoaded } from "../../utils/Util";
 onlyOn(isEnvironmentType(EnvType.Mocked), () => {
   describe("Edit Smart Event Instance Tests", () => {
     beforeEach(() => {
-      cy.visit("/");
+      //Open 'Instance one' page
+      cy.visit("/instance/3543edaa-1851-4ad7-96be-ebde7d20d717");
       pageWasLoaded();
     });
 
     it("Edit Error handler configuration", () => {
-      //Open 'Instance one' page
-      cy.ouiaId("Instances list table", "PF4/Table")
-        .ouiaId("Instance one", "PF4/TableRow")
-        .within(() => {
-          cy.get(
-            "td:first a[data-testid='tableInstances-linkInstance']"
-          ).click();
-        });
-
       // Switch to 'Error handling' tab
       cy.ouiaId("error-handling", "PF4/TabButton").click();
       cy.ouiaId("error-handling-section", "PF4/Text").should("be.visible");
@@ -52,15 +44,6 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
     });
 
     it("Switch from 'Webhook' to 'Ignore' Error handler", () => {
-      //Open 'Instance one' page
-      cy.ouiaId("Instances list table", "PF4/Table")
-        .ouiaId("Instance one", "PF4/TableRow")
-        .within(() => {
-          cy.get(
-            "td:first a[data-testid='tableInstances-linkInstance']"
-          ).click();
-        });
-
       // Switch to error handling tab
       cy.ouiaId("error-handling", "PF4/TabButton").click();
 
--- a/cypress/integration/instance/InstanceEdit.ts
+++ b/cypress/integration/instance/InstanceEdit.ts
@@ -7,11 +7,12 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
       //Open 'Instance one' page
       cy.visit("/instance/3543edaa-1851-4ad7-96be-ebde7d20d717");
       pageWasLoaded();
+
+      // Switch to error handling tab
+      cy.ouiaId("error-handling", "PF4/TabButton").click();
     });
 
     it("Edit Error handler configuration", () => {
-      // Switch to 'Error handling' tab
-      cy.ouiaId("error-handling", "PF4/TabButton").click();
       cy.ouiaId("error-handling-section", "PF4/Text").should("be.visible");
 
       // Do simple change in 'Error handling' conf
@@ -44,9 +45,6 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
     });
 
     it("Switch from 'Webhook' to 'Ignore' Error handler", () => {
-      // Switch to error handling tab
-      cy.ouiaId("error-handling", "PF4/TabButton").click();
-
       // Wait tab is loaded
       cy.get(".pf-c-description-list__term").should("have.length", 5);
       // Once tab is loaded start edit mode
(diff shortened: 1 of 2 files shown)


cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +50,67 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
+      cy.get("button.pf-c-select__menu-item", { timeout: 10000 })
+        .contains("Ignore")
+        .click();
+      cy.ouiaId("submit", "PF4/Button").click();
+
+      cy.ouiaId("Instances list table", "PF4/Table")
+        .ouiaId("Instance one", "PF4/TableRow")
+        .should("be.visible")
+        .within(() => {
+          //The user cannot go to the instance's details if it is not ready
```
vpellegrino, line 91: FYI this is going to change with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/243

vpellegrino (approved):
This test looks good to me. Just note one behavior (inner comment) that will change soon.

cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +50,67 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
             });
         });
     });
+
+    it.only("Switch from 'Webhook' to 'Ignore' Error handler", () => {
```
kelvah, line 54: Is this `it.only` intentional?

kelvah (approved):
LGTM

cypress/integration/instance/InstanceEdit.ts
jomarko: no, it is a mistake. I will remove it, thank you

cypress/integration/instance/InstanceEdit.ts
jomarko: I will remove that comment then probably

cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +50,65 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
+      cy.ouiaId("error-handling", "PF4/TabButton").click();
+
+      // Wait tab is loaded
+      cy.get(".pf-c-description-list__term").should("have.length", 5);
+      // Once tab is loaded start edit mode
+      cy.ouiaId("error-handling", "PF4/TabContent").within(() => {
+        cy.ouiaId("edit", "PF4/Button")
+          .should("have.attr", "aria-disabled")
+          .and("equal", "false");
+        cy.ouiaId("edit", "PF4/Button").click();
```
sjamboro, line 74: @jomarko is possible that the edit button is disabled when the page is loaded? Do you expect DOM exception that you repeat `ouiaId("edit", "PF4/Button")`?

cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +50,65 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
+      });
+
+      // Change error handling strategy to 'Ignore'
+      cy.ouiaId("error-handling-method-selector", "PF4/Select").within(() => {
+        cy.get("button").should("not.have.attr", "disabled");
+      });
```
sjamboro, line 80: Similar question. Is there any reason for using within? This construction should work:
```
cy.ouiaId("error-handling-method-selector", "PF4/Select").find("button").should("not.have.attr", "disabled");
```

cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +50,65 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
+      cy.ouiaId("error-handling-method-selector", "PF4/Select").click();
+      cy.get("button.pf-c-select__menu-item", { timeout: 10000 })
```
sjamboro, line 82: In case that you want avoid this selector and use ouiaId. You can define it here:
https://github.com/jomarko/sandbox-ui/blob/MGDOBR-1095-test/src/app/Instance/ErrorHandling/ErrorHandlingSelection.tsx#L60-L63
https://github.com/jomarko/sandbox-ui/blob/MGDOBR-1095-test/src/app/Instance/ErrorHandling/ErrorHandlingSelection.tsx#L72-L74

These function components are not OUIA compilant so you have to use whole names of OUIA attributes (`data-ouia-component-id` and `data-ouia-component-type`).

cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +50,65 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
+              cy.ouiaId("creating", "PF4/Button").should(
+                "have.text",
+                "Creating"
+              );
+              cy.ouiaId("ready-shortly", "QE/HelperTextItem").should(
+                "have.text",
+                "This will be ready shortly."
+              );
+            });
+          cy.ouiaId("creating", "PF4/Button").click();
```
sjamboro, line 106: Do we need to test Popover here? Is possible to check the `ouiaId("ready", "QE/ResourceStatus")` status

cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +50,65 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
         });
     });
+
+    it("Switch from 'Webhook' to 'Ignore' Error handler", () => {
+      //Open 'Instance one' page
+      cy.ouiaId("Instances list table", "PF4/Table")
+        .ouiaId("Instance one", "PF4/TableRow")
+        .within(() => {
+          cy.get(
+            "td:first a[data-testid='tableInstances-linkInstance']"
```
sjamboro, line 60: @jomarko All tests in this test suite are running with same instance. You can open directly the instance page in `beforeEach`.
See: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/2c2203e90dbc1a1b633a788ec7afd8700bf60320/cypress/integration/instance/InstancePage.ts#L19

sjamboro (approved):
@jomarko thank you for your commit. Let me share my observation. Feel free to implement them or merge the current version.

cypress/integration/instance/InstanceEdit.ts
jomarko: yes the edit button is disabled when page is loaded. that is why I have this line in test
```
        cy.ouiaId("edit", "PF4/Button")
          .should("have.attr", "aria-disabled")
          .and("equal", "false");
```

That line above ensures I do not click on button soooner that in is enabled.

Click on disabled button was not throwing any error, but next lines of test were failing, because expected form fields/elements, were not present in the web page.

cypress/integration/instance/InstanceEdit.ts
jomarko: well, `within` is my personal preference, is there any cypress recommendation/good practice to use `find` instead of `within`?

cypress/integration/instance/InstanceEdit.ts
jomarko: thank you for a tip, will do in a separate commit

jomarko: see 4b871ec

cypress/integration/instance/InstanceEdit.ts
jomarko: thank you for a comment. is add5500 what you had in mind @sjamboro ?

cypress/integration/instance/InstanceEdit.ts
jomarko: thank you for a tip @sjamboro , is a0781ad what you had in mind please?

cypress/integration/instance/InstanceEdit.ts
sjamboro: thank you for your answer

cypress/integration/instance/InstanceEdit.ts
sjamboro: I do not think so. It surprised me because I usually use `find` when I search 1 element and `within` when I need to assert more than 1 element.

cypress/integration/instance/InstanceEdit.ts
sjamboro: @jomarko yes

sjamboro (approved):
LGTM ... @jomarko thank you for your PR

cypress/integration/instance/InstanceEdit.ts
```diff
@@ -50,5 +42,39 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
             });
         });
     });
+
+    it("Switch from 'Webhook' to 'Ignore' Error handler", () => {
+      // Switch to error handling tab
+      cy.ouiaId("error-handling", "PF4/TabButton").click();
```
sjamboro, line 48: It depends on you but this line is same for both tests. It could be move to `beforeEach` method.

jomarko: pushed in a separate commit 1400d3c

jomarko:
`InstanceEdit` passed as the only touched test file.

The test failures in the last CI run are all caused by cookies consent dialogue. I am merging this PR.

PR #207, opened 2022-09-30
MGDOBR-1095: "Create SE Instance" form validation issue with error handling selection

vpellegrino:
Task: https://issues.redhat.com/browse/MGDOBR-1095

### DIFF
--- a/src/app/Instance/CreateInstance/CreateInstance.tsx
+++ b/src/app/Instance/CreateInstance/CreateInstance.tsx
@@ -432,6 +432,7 @@ const CreateInstance = (props: CreateInstanceProps): JSX.Element => {
               isDisabled={formIsDisabled}
               onMethodSelection={(errorMethod: string): void => {
                 setErrorHandlingParameters({});
+                setErrorHandlingSchema({});
                 if (errorMethod === ERROR_HANDLING_METHODS.default.value) {
                   setErrorHandlingSchemaId(null);
                 } else {


jomarko (approved):
No manual check done. Will see if that works during my work on #203 . Code change seems to be reasonable, however I do not have deep knowledge of the codebase.

kelvah (commented):
Hi @vpellegrino, thank you for the fix. 

I've manually checked and the "locked" form is gone :+1: 

I am struggling to understand though how setting the schema to an empty object is fixing it. Could you clarify?

Sorry, I am probably missing something here. In my defense it's Friday afternoon.

vpellegrino:
> Hi @vpellegrino, thank you for the fix.
> 
> I've manually checked and the "locked" form is gone 👍
> 
> I am struggling to understand though how setting the schema to an empty object is fixing it. Could you clarify?
> 
> Sorry, I am probably missing something here. In my defense it's Friday afternoon.

I checked that setting the schema to a new object, triggers the validation function to be executed.

When you switch from webhook to kafka topic, the schema is populated by the API response, so the validation gets reset. In case of none, no API call is performed, so we need to manually force a state change for such variable.

kelvah:
> I checked that setting the schema to a new object, triggers the validation function to be executed.
> 
> When you switch from webhook to kafka topic, the schema is populated by the API response, so the validation gets reset. In case of none, no API call is performed, so we need to manually force a state change for such variable.

Thanks for the tips :+1: 

Sorry but I've only realized now that I did something too to solve this inside #193.
It's here: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/7a01fbf63812361521d8a01ef4ca2b41db6562b5/src/app/Instance/CreateInstance/components/ErrorHandling.tsx#L50-L57

Wdyt?
Different approach of course. Maybe more explicit?

vpellegrino:
> > I checked that setting the schema to a new object, triggers the validation function to be executed.
> > When you switch from webhook to kafka topic, the schema is populated by the API response, so the validation gets reset. In case of none, no API call is performed, so we need to manually force a state change for such variable.
> 
> Thanks for the tips 👍
> 
> Sorry but I've only realized now that I did something too to solve this inside #193. It's here:
> 
> https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/7a01fbf63812361521d8a01ef4ca2b41db6562b5/src/app/Instance/CreateInstance/components/ErrorHandling.tsx#L50-L57
> 
> Wdyt? Different approach of course. Maybe more explicit?

You raised this defect while you were developing to the instance creation refactor (the code you are referring to).
Are you sure that it fixes all scenarios? If yes, then we can ignore this PR.

vpellegrino:
Closing as per comments

PR #215, opened 2022-10-07
MGDOBR-1133: [UI] SSO shows a cookie's popover

sjamboro:
https://issues.redhat.com/browse/MGDOBR-1133

This PR presents the hot fix: check if the popover is presented and if yes than click at the 'Accept default' button.


kelvah (approved):
Thanks @sjamboro for the quick fix. LGTM

PR #217, opened 2022-10-07
NO-JIRA: remove unused @types/node dependency

kelvah:
It was introduced (by me actually) with #24 but it's not required for msw.


manstis (approved):

sjamboro (approved):

vpellegrino (approved):
LGTM

PR #223, opened 2022-10-11
MGDOBR-1133: Fix timeouts

sjamboro:

PR #226, opened 2022-10-14
[No-Jira] Accept missing instance progress poppup as valid state of t…

jomarko:
…he application

There are some non deterministic situations in our application, when some popover dialog may or may not be present on the page. Examples of such situations are:
1.) Consent with cookies popover in the very start of the application/test 2.) Smart Event Instance status progress popover

The 1.) is for some user displayed, for some it is not. We Are not sure why this happens. The guess for a root case is a country of the machine, where the test is running. Needs more investigations/test results tracking. The 2.) displays progress as three steps. However the time combination of time needed for the operation on backend with polling interval from UI causes, soemtimes the progress update is not 'smooth' and popover disappears faster than expected in the test.

*This PR tries to stabilize 2.)*

The error handling was addopted according to th ecyprtess example:
- https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/fundamentals__errors/cypress/e2e/test-fails.cy.js

Screenshot from 2022-10-14 10-19-00

### DIFF
--- a/cypress/integration/instance/InstanceCreation.ts
+++ b/cypress/integration/instance/InstanceCreation.ts
@@ -69,12 +69,13 @@ describe("the 'Create a SE instance' Modal", () => {
           console.error("runnable", runnable);
 
           if (
+            isEnvironmentType(EnvType.Dev) &&
             e.name === "AssertionError" &&
             e.message.includes("of 3 steps completed")
           ) {
-            return false;
+            return true; //does not matter if true or false
           }
-          return true;
+          throw e;
         });
 
         progressStepsStatuses(SEInstanceStatus.ACCEPTED);
--- a/cypress/integration/instance/InstanceCreation.ts
+++ b/cypress/integration/instance/InstanceCreation.ts
@@ -59,6 +59,24 @@ describe("the 'Create a SE instance' Modal", () => {
           "have.text",
           "0 of 3 steps completed"
         );
+
+        // Prepare cypress for 'non deterministic' updates in 'Status' popover
+        // The 'deterministic' updates would be: (0 of 3, 1 of 3, 2 of 3 and Done)
+        // However in reality it may happen: (0 of 3, 1 of 3 and Done) or (0 of 3, 2 of 3 and Done) or ...
+        // see https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/fundamentals__errors/cypress/e2e/test-fails.cy.js
+        cy.on("fail", (e, runnable) => {
+          console.error("error", e);
+          console.error("runnable", runnable);
+
+          if (
+            e.name === "AssertionError" &&
+            e.message.includes("of 3 steps completed")
+          ) {
+            return false;
+          }
+          return true;
+        });
+
         progressStepsStatuses(SEInstanceStatus.ACCEPTED);
         cy.ouiaId("steps-count", "QE/StackItem", { timeout: 120000 }).should(
           "have.text",


jomarko:
Please do not merge this PR, I want to rerun it several times to see if my code changes works

jomarko:
Attempt No#1 passed

jomarko:
Attempt No#2 failed due to non deterministic cookies consent:
the 'Create a SE instance' Modal -- Submit -- before each hook (failed)

jomarko:
Attempt No#3 passed.

jomarko:
Attempt No#4 passed

jomarko:
Attempt No#5 failed. All failures due tot he same reason as https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/226#issuecomment-1278867030

jomarko:
Attempt No#6 passed

jomarko:
Attempt No#7 failed with random failure. 
the 'Create a SE instance' Modal -- Cancel -- before each hook (failed)

jomarko:
Attempt No#8 failed due to some longer existence of deleted row.

UI react on the REST interactions -- Data are visible -- after all hook (failed)

jomarko:
Attempt No#9 failed due to cookies consent, same reason as No#5 and No#2

cypress/integration/instance/InstanceCreation.ts
```diff
@@ -59,6 +59,24 @@ describe("the 'Create a SE instance' Modal", () => {
...
+
+          if (
+            e.name === "AssertionError" &&
+            e.message.includes("of 3 steps completed")
+          ) {
+            return false;
+          }
+          return true;
+        });
+
```
kelvah, line 79: This is a neat solution :+1: 

I was wondering if we could have also disabled this "steps" checks for dev with the `onlyOn` approach. Wdyt?

kelvah (approved):
LGTM :+1: 
Thanks @jomarko. I've left a comment but it's only a question.

cypress/integration/instance/InstanceCreation.ts
jomarko: @kelvah thank you, I am testing your suggestion https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/230, again there, I will run 10 attempts.

vpellegrino (approved):
LGTM

cypress/integration/instance/InstanceCreation.ts
sjamboro: Is possible to return true in the `.on("fail",...)` code part when the test is running on Mocked Env? I mean add this condition: `if (isEnvironmentType(EnvType.Dev))`

jomarko: Not sure @sjamboro I understand what do you mean. Do you mean the test should always fail, if this 'special handling of assertion error' happens on mocked environment?

Maybe you can share whole `cy.on("fail" ...` block you have in mind?

sjamboro:
> Attempt No#8 failed due to some longer existence of deleted row.
> 
> UI react on the REST interactions -- Data are visible -- after all hook (failed)

It should be solved by #223.

cypress/integration/instance/InstanceCreation.ts
sjamboro: I am working on that I have tried to simulate the error. I have rewritten the expectation like `4 of 3 steps completed` and I do not get the result as I expected.

sjamboro: @jomarko do you try to simulate the error locally? I deleted the condition and it seems that the test is allways marked as successfull (no metter if I return `true` or `false`). See:

```
       cy.on("fail", (e, runnable) => {
          return true;
        }); 

        cy.ouiaId("steps-count", "QE/StackItem", { timeout: 10000 }).should(
          "have.text",
          "4 of 3 steps completed"
        );
```

sjamboro: @jomarko I have included the code which has the expected behaviour (IIUC).

**In case I assert:** `cy.ouiaId("steps-count", "QE/StackItem", { timeout: 10000 }).should("have.text", "4 of 3 completed");`
A) **Mask error on DEV**
- the console contains the error log but the test is marked as succeeded.
- I am not sure if `console.error(...);` works here even if I delete them the console contains the error log and there is no double log.

B) **failed on MOCKED**
- the AssertionError alert is present and the test is marked as failed.

C) **the condition masked some problem**
- the whole message is `to have text 4 of 3 completed, but the text was 0 of 3 steps completed`
- the important part is that we assert `.should('have.text')` so if the element has text which includes `of 3 steps completed` the whole test is marked as succeeded

```
        cy.on("fail", (e, runnable) => {
          console.error("error", e);
          console.error("runnable", runnable);
      
          if (
            isEnvironmentType(EnvType.Dev) &&
            e.name === "AssertionError" &&
            e.message.includes("of 3 steps completed")
          ) {
            return true; //does not matter if true or false
          }
          throw e;
        });
```

sjamboro (commented):
As I mentioned above ... I have tried the code locally and it does not work as expected.

cypress/integration/instance/InstanceCreation.ts
jomarko: @sjamboro thank you for an investigation. I pushed your proposal in a separate commit.

sjamboro (approved):

PR #228, opened 2022-10-14
MGDOBR-1129: Refactor the Instance Page layout for adherence with UX prototype

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1129

This PR implements the "full height layout with fixed actions" for the error handling tab of bridges.

I did some investigation before starting working on it. 

PF offers a "sticky actions" solution with the `PageSection` component ([example](https://www.patternfly.org/v4/components/page/html-demos/#sticky-section-bottom)). In order to use it though, the involved `PageSection`'s need to be all at the same _root_ level. In our case, the EH actions are deep down in the tree from where the page layout resides (`InstancePage` :arrow_right: `ErrorHandlingTabContent` :arrow_right: `ErrorHandlingEdit`). Another hurdle comes with the presence of PF `Tabs` within `InstancePage`, because each `Tab` interfere with [filled](https://www.patternfly.org/v4/components/page#with-or-without-fill) `PageSection`'s behavior.

The only option I saw was a "custom" layout implementation. We did a similar thing for the "Processor edit" page, where actions are fixed at the bottom of the viewport. I opted for extracting this layout into a separate component, `StickyActionsLayout`, and reuse it for the EH too.

Doing so I had to change a bunch of things around unfortunately. I'll add some review comments to describe what's happening where.

I'm not particularly happy about the amount of changes I had to make, some of them not so self-explanatory. 

Let me know what you think.

### DIFF
--- a/src/app/Instance/ErrorHandling/ErrorHandlingDetail.tsx
+++ b/src/app/Instance/ErrorHandling/ErrorHandlingDetail.tsx
@@ -5,7 +5,6 @@ import {
   DescriptionListDescription,
   DescriptionListGroup,
   DescriptionListTerm,
-  PageSection,
   PageSectionVariants,
   Skeleton,
   Split,
@@ -23,6 +22,7 @@ import {
   getErrorHandlingMethodByType,
 } from "../../../types/ErrorHandlingMethods";
 import ConfigParameters from "@app/components/ConfigParameters/ConfigParameters";
+import ErrorHandlingPageSection from "@app/Instance/ErrorHandling/ErrorHandlingPageSection";
 
 export interface ErrorHandlingDetailProps {
   errorHandlingType?: string;
@@ -78,11 +78,7 @@ export const ErrorHandlingDetail = ({
   }, [errorHandlingParameters]);
 
   return (
-    <PageSection
-      variant={PageSectionVariants.light}
-      isFilled
-      className={"pf-u-h-100"}
-    >
+    <ErrorHandlingPageSection variant={PageSectionVariants.light}>
       <Stack hasGutter>
         <StackItem>
           <Split hasGutter>
@@ -146,6 +142,6 @@ export const ErrorHandlingDetail = ({
           </DescriptionList>
         </StackItem>
       </Stack>
-    </PageSection>
+    </ErrorHandlingPageSection>
   );
 };
--- a/src/app/Instance/ErrorHandling/ErrorHandlingDetail.tsx
+++ b/src/app/Instance/ErrorHandling/ErrorHandlingDetail.tsx
@@ -1,10 +1,18 @@
 import {
   Alert,
+  Button,
   DescriptionList,
   DescriptionListDescription,
   DescriptionListGroup,
   DescriptionListTerm,
   Skeleton,
+  Split,
+  SplitItem,
+  Stack,
+  StackItem,
+  Text,
+  TextContent,
+  TextVariants,
 } from "@patternfly/react-core";
 import React, { useMemo } from "react";
 import { useTranslation } from "@rhoas/app-services-ui-components";
@@ -21,6 +29,8 @@ export interface ErrorHandlingDetailProps {
   schema?: object;
   isSchemaLoading: boolean;
   apiError?: string;
+  onEdit: () => void;
+  isEditDisabled: boolean;
 }
 
 export const ErrorHandlingDetail = ({
@@ -30,6 +40,8 @@ export const ErrorHandlingDetail = ({
   schema,
   isSchemaLoading,
   apiError,
+  onEdit,
+  isEditDisabled,
 }: ErrorHandlingDetailProps): JSX.Element => {
   const { t } = useTranslation(["openbridgeTempDictionary"]);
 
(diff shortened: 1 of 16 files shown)


src/app/Instance/InstancePage/InstancePage.tsx
```diff
@@ -272,16 +264,40 @@ const InstancePage = (): JSX.Element => {
...
-                  bridge={bridge}
-                  isBridgeLoading={isBridgeLoading}
-                />
-              </PageSection>
-            </Tab>
+            />
           </Tabs>
         </PageSection>
+        <PageSection isFilled padding={{ default: "noPadding" }}>
+          <TabContent
```
kelvah, line 271: Switching to the separate content tabs setup, where tabs content is not nested inside tab links ([docs example](https://www.patternfly.org/v4/components/tabs#separate-content)).
I needed this to explicitly use `TabContent` and modify it a bit.

src/app/Instance/InstancePage/InstancePage.tsx
```diff
@@ -272,16 +264,40 @@ const InstancePage = (): JSX.Element => {
...
+              />
+            )}
+          </TabContent>
+          <TabContent
+            key={INSTANCE_PAGE_TAB_KEYS["error-handling"]}
+            eventKey={INSTANCE_PAGE_TAB_KEYS["error-handling"]}
+            id={"instance-page__tabs-error-handling"}
+            ouiaId="error-handling"
+            activeKey={activeTabKey}
+            className={"pf-c-page__main-section pf-u-h-100"}
```
kelvah, line 291: Altering `TabContent` styles to make its content fill the available page height. 

Adding the `pf-c-page__main-section` to make it behave like a `PageSection`'s, and `pf-u-h-100`, a PF css utility class to set the height to `100%`

src/app/Instance/InstancePage/ErrorHandlingTabContent.tsx
```diff
@@ -167,62 +155,32 @@ export const ErrorHandlingTabContent = ({
...
+          errorHandlingType={bridge?.error_handler?.type}
+          errorHandlingParameters={bridge?.error_handler?.parameters}
+          isBridgeLoading={isBridgeLoading}
+          isSchemaLoading={isSchemaLoading}
+          schema={currentSchema}
+          apiError={apiError}
+          onEdit={(): void => setIsEditing(true)}
+          isEditDisabled={editIsDisabled}
+        />
+      )}
```
kelvah, line 183: Separating the layout of the EH tab between the "view" and "edit" mode.
The PF `Stack` layout was mainly useful only to the view mode.

src/app/Instance/ErrorHandling/ErrorHandlingEdit.tsx
```diff
@@ -71,55 +74,71 @@ export const ErrorHandlingEdit = ({
...
-        registerValidation={registerValidateParameters}
-        onChange={onErrorHandlingParametersChange}
-        parameters={errorHandlingParameters}
-      />
-      <PageSection
-        stickyOnBreakpoint={{ default: "bottom" }}
-        padding={{ default: "noPadding" }}
-        style={{ boxShadow: "none" }}
+      <StickyActionsLayout
+        actions={
```
kelvah, line 84: Adding the new layout component and passing if the EH form and form actions as props,

src/app/Processor/ProcessorEdit/ProcessorEdit.tsx
```diff
@@ -58,6 +56,7 @@ import { ActionModal } from "@app/components/ActionModal/ActionModal";
 import { css } from "@patternfly/react-styles";
 import { HelpIcon } from "@patternfly/react-icons";
 import { ExternalLink } from "@rhoas/app-services-ui-components";
+import StickyActionsLayout from "@app/components/StickyActionsLayout/StickyActionsLayout";
```
kelvah, line 59: Lots of changes in this file but mainly removing the layout code that now resides inside `StickyActionsLayout`, and replacing it with the new component.

src/app/Instance/ErrorHandling/ErrorHandlingDetail.tsx
```diff
@@ -21,6 +31,8 @@ export interface ErrorHandlingDetailProps {
   schema?: object;
   isSchemaLoading: boolean;
   apiError?: string;
+  onEdit: () => void;
+  isEditDisabled: boolean;
```
kelvah, line 35: `ErrorHandlingDetail` now renders the "Edit" button, previously located inside `ErrorHandlingTabContent`.
Adding some props to plug it to the parent component.

src/app/Processor/ProcessorEdit/ProcessorEdit.css
```diff
@@ -1,35 +1,6 @@
...
-  /*position: static;*/
-  min-height: 0;
-}
-
-.processor-edit__content-wrap {
-  overflow-x: hidden;
-  overflow-y: auto;
-  word-break: break-word;
-}
-
```
kelvah, line 29: Old sticky actions layout used within `ProcessorEdit`, now made generic and moved to the new layout component.

manstis (approved):
I didn't drill down into the new layout code a great deal.

However I ran it; things look OK and work OK, so OK with me :+1:

jomarko (commented):
Thank you for a PR @kelvah some comments form me inline.

vpellegrino (approved):
That's a huge improvement @kelvah 
I appreciate the re-usage of `StickyActionsLayout` in both `ProcessorEdit` and `ErrorHandlingEdit`.

src/app/Instance/InstancePage/ProcessorsTabContent.tsx
```diff
@@ -189,40 +189,40 @@ export const ProcessorsTabContent = ({
...
+            pageNumber={currentPage}
+            pageSize={currentPageSize}
+            totalRows={totalRows ?? 0}
+            renderActions={({ row, ActionsColumn }): JSX.Element => (
+              <ActionsColumn items={tableActions(row)} />
+            )}
+          >
+            <EmptyState variant="large">
+              <EmptyStateIcon icon={PlusCircleIcon} />
+              <Title headingLevel="h4" size="lg">
```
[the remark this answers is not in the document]
kelvah: I didn't change this :slightly_smiling_face: 

Anyway custom `Title` sizes are legit. [See PF docs](https://www.patternfly.org/v4/components/title/#default-sizes). 
The default sizes can be changed.
One thing is semantic value of the title element, the heading level, another one is the styling and how it should look like, the sizing. They are related but not strictly.

I would anyway argue that the `h4` level is wrong here. The `no processors` message should probably be the second in importance within the page, after the `h1` of the instance name. 

I'll change it to `h2` with `lg` size if you agree. @vpellegrino wdyt?

src/app/Instance/ErrorHandling/ErrorHandlingDetail.tsx
```diff
@@ -64,36 +78,74 @@ export const ErrorHandlingDetail = ({
...
-          title={apiError}
-          aria-live="polite"
-          isInline
-        />
-      )}
-    </DescriptionList>
+    <PageSection
+      variant={PageSectionVariants.light}
+      isFilled
+      className={"pf-u-h-100"}
```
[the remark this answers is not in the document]
kelvah: Your comment made me think. I guess we could also improve things a bit more.

For the new layout to work, all the page sections inside the error handling area must have both `isFilled` and `height: 100%`. Changing this configuration in any of them will break something. I think this is not so explicit when looking at the code cause it's split across different files.

I moved these page components to a separate one (`ErrorHandlingPageSection`) to make things easier to understand.

About having an enum to store the value of `pf-u-h-100`, I think it's not so worth. It doesn't guarantee that the class will always exist (it's a CSS utility class [provided by PF](https://www.patternfly.org/v4/utilities/sizing#usage-1)) nor that it will do its work. It will help making sure we write it right, but at the cost of remembering how the enum is named. I think it's easier to just refer to PF docs when in need.

jomarko (approved):
Thank you @kelvah .

Approving. The current test failure seems to be one, that I try to resolve in #226

```
This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add &apos;guards&apos; which delay Cypress from running new commands.

https://on.cypress.io/element-has-detached-from-dom" type="CypressError"><![CDATA[CypressError: Timed out retrying after 120000ms: `cy.should()` failed because this element is detached from the DOM.

`<div data-ouia-component-id="steps-count" data-ouia-component-type="QE/StackItem" class="pf-l-stack__item">1 of 3 ...</div>`

Cypress requires elements be attached in the DOM to interact with them.

The previous command that ran was:

  > `cy.ouiaId()`

This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.
```

PR #118, opened 2022-08-03
MGDOBR-523: Bridges/Processors - New Status label component

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-523

This PR is dependent on:
- https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1033

Replacing the temporary label with a new component inspired by the kafka status label poc.

The main difference is that now creation statuses (`accepted`, `preparing`, `provisioning`) are grouped under the `Creating` label. Same for the deleting statuses.

Also, clicking on the `Creating` label will open a popover with more information.

PR #229, opened 2022-10-18
MGDOBR-1148: Refresh processor detail information during creation/update process

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1148

### DIFF
--- a/src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
+++ b/src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
@@ -45,6 +45,7 @@ import {
 import { APIErrorCodes } from "@openapi/generated/errors";
 import { ActionModal } from "@app/components/ActionModal/ActionModal";
 import SEStatusLabel from "@app/components/SEStatusLabel/SEStatusLabel";
+import { usePolling } from "../../../hooks/usePolling/usePolling";
 
 const ProcessorDetailPage = (): JSX.Element => {
   const { instanceId, processorId } = useParams<ProcessorRouteParams>();
@@ -57,6 +58,7 @@ const ProcessorDetailPage = (): JSX.Element => {
   );
 
   const [isEditing, setIsEditing] = useState(false);
+  const [processorRefreshInterval, setProcessorRefreshInterval] = useState(0);
   const [currentProcessor, setCurrentProcessor] = useState<ProcessorResponse>();
   const [isActionsOpen, setIsActionsOpen] = useState(false);
   const [existingProcessorName, setExistingProcessorName] = useState<
@@ -102,6 +104,11 @@ const ProcessorDetailPage = (): JSX.Element => {
     getProcessor(instanceId, processorId);
   }, [getProcessor, instanceId, processorId]);
 
+  usePolling(
+    () => getProcessor(instanceId, processorId),
+    processorRefreshInterval
+  );
+
   const {
     updateProcessor,
     processor: updatedProcessor,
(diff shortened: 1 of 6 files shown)


jomarko (approved):
Changes looks fine for me, however we need to update ProcessorTest ASAP, so be aware of changes coming. https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/232

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
```diff
@@ -119,11 +126,17 @@ const ProcessorDetailPage = (): JSX.Element => {
 
   useEffect(() => {
     setCurrentProcessor(processor);
+    if (processor && !canEditResource(processor?.status)) {
+      setProcessorRefreshInterval(refreshInterval);
+    } else {
+      setProcessorRefreshInterval(0);
```
manstis, line 132: IDK if I misunderstand.. 

when you want to clear the polling should you not also call the `clearInterval()` [method](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/main/src/hooks/usePolling/usePolling.ts#L20) returned from `usePolling(..)`? Or, should the call to `usePolling(..)` have a dependency on `refreshInterval` to ensure it is set to `0`?

vpellegrino (approved):
It looks good to me.
Thank you @kelvah

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
kelvah: The way to stop `usePolling` is to pass it a delay of `0`.

https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/b919ee59566392bad320630c250238ffdd693bae/src/hooks/usePolling/usePolling.ts#L13-L21

When doing so, its `useEffect` will run again because it depends on this delay value. 

Before the new render, `useEffect` will run its cleanup function, the one it's being returned inside it, clearing the existing `setInterval`.

On the re-render, a new `setInterval` will not be set because the delay is `0`.

(You can find more about this mechanism [here](https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1), especially in the last paragraph)

I'm [passing](https://github.com/kelvah/sandbox-ui/blob/fbbf0d706ec42e88a54467aa6ad2fce0ab23cd21/src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx#L107-L110) `processorRefreshInterval` to `usePolling`, and setting It to `0` or to `refeshInterval` when needed.
`refreshInterval` is only a constant to store the delay value in a single place.

Going back to your original question, `usePolling` does not return anything! You may ask why then its return type is `void | (() => void)`. That's clearly a mistake :)

I introduced it when working on #118, but I got things wrong. The polling within [`useResourceDelayedStatus`](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/b919ee59566392bad320630c250238ffdd693bae/src/app/components/SEStatusLabel/useResourceDelayedStatus.ts) was not cleared at all. Fixed now.

Thanks @manstis for bringing this up.

src/app/components/SEStatusLabel/SEStatusLabel.tsx
```diff
@@ -15,20 +15,10 @@ export interface SEStatusLabelProps {
   resourceType: "bridge" | "processor";
   /** When the resource was requested (created or updated) */
   requestedAt: Date;
-  /** Flag to decide if the delayed warning check should run only one time.
-   *  If set to "false", or not set, the component will check every 1s if the
-   *  delayed message should be displayed.
-   */
-  singleDelayedCheck?: boolean;
```
kelvah, line 22: Removing this feature from the status label component. It was specifically added to make it work within the processor page, where the data was not refreshed. Now it's not necessary anymore. (If you need more info, [here's a discussion](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/118#discussion_r937813702) about it)

src/app/Processor/ProcessorDetailPage/ProcessorDetailPage.tsx
manstis: Thank-you for the detailed explanation.

jomarko (approved):
Manually it seems to work. Thank you @kelvah

manstis (approved):
Thank-you @kelvah LGTM :+1:

PR #230, opened 2022-10-18
[No-Jira] Skip checks of 'non deterministic' updates in 'Status' popo…

jomarko:
…ver in Dev environment

This is an alternative for #226

### DIFF
--- a/cypress/integration/instance/InstanceCreation.ts
+++ b/cypress/integration/instance/InstanceCreation.ts
@@ -47,31 +47,37 @@ describe("the 'Create a SE instance' Modal", () => {
         cy.ouiaId("creating", "PF4/Button").click();
       });
 
-    //Assert that the process of the instnace's creation is monitored
-    cy.ouiaId("se-status", "QE/Popover")
-      .should("be.visible")
-      .within(() => {
-        cy.ouiaType("QE/StackItem").should("have.length", "3");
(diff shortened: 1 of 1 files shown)


jomarko:
Attempt No#1 PASSED.

jomarko:
Attempt No#3 FAILED the login phase.
the 'Create a SE instance' Modal -- Submit and expect error -- before each hook (failed)

jomarko:
Attempt No#4 PASSED.

jomarko:
Attempt No#5 PASSED.

kelvah (approved):
LGTM :ok_hand: 
Thanks @jomarko for considering feedback and for taking care of exploring alternative solutions. Great job!

sjamboro (commented):
In this case I miss validation that the popover is present on the dev and contains all appropriate states. I think that the solution from #226 is better for me. 

The approach of #226 has these benefits:
- it tries to validate all steps on dev and mocked environment
- it defines a condition when the environment is not deterministic
- IMO it is more readable than definition what can run on DEV and what will run on MOCKED

The approach of #226 has this negative:
- we need to define the condition well to not mask the real problem

jomarko:
Taking into account https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/230#pullrequestreview-1149098547

I will close #230

PR #232, opened 2022-10-18
Update ProcessorTest

jomarko:
This is a follow up PR:
- for https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/229/files
- and https://issues.redhat.com/browse/MGDOBR-797

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "trailingComma": "es5",
   "singleQuote": false,
-  "semi": true
+  "semi": true,
+  "quoteProps": "consistent"
 }
--- a/cypress/integration/ProcessorTest.ts
+++ b/cypress/integration/ProcessorTest.ts
@@ -39,19 +39,19 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
         cy.ouiaId("add-filter", "PF4/Button").should("be.visible").click();
 
         cy.ouiaId("item-0").within(() => {
-          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[0].Key);
-          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[0].Type);
+          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[0]["Key"]);
+          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[0]["Type"]);
         });
         cy.ouiaId("item-0") //The filter-value was detached from DOM and we need to find the context again.
           .ouiaId("filter-value", "PF4/TextInput")
-          .type(filters[0].Value);
+          .type(filters[0]["Value"]);
         cy.ouiaId("item-1").within(() => {
-          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[1].Key);
-          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[1].Type);
+          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[1]["Key"]);
+          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[1]["Type"]);
         });
         cy.ouiaId("item-1") //The filter-value was detached from DOM and we need to find the context again.
           .ouiaId("filter-value", "PF4/TextInput")
-          .type(filters[1].Value);
+          .type(filters[1]["Value"]);
 
         //Transformation
         cy.ouiaId("transformation-description", "PF4/Text")
@@ -159,21 +159,21 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
         //Filters
         cy.ouiaId("add-filter", "PF4/Button").should("be.visible").click();
         cy.ouiaId("item-0").within(() => {
-          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[0].Key);
-          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[0].Type);
+          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[0]["Key"]);
+          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[0]["Type"]);
         });
         //The filter-value was detached from DOM and we need to find the context again.
         cy.ouiaId("item-0")
           .ouiaId("filter-value", "PF4/TextInput")
-          .type(filters[0].Value);
+          .type(filters[0]["Value"]);
         cy.ouiaId("item-1").within(() => {
-          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[1].Key);
-          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[1].Type);
+          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[1]["Key"]);
+          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[1]["Type"]);
         });
         //The filter-value was detached from DOM and we need to find the context again.
         cy.ouiaId("item-1")
           .ouiaId("filter-value", "PF4/TextInput")
-          .type(filters[1].Value);
+          .type(filters[1]["Value"]);
 
         cy.ouiaId("submit").click();
 
@@ -241,20 +241,20 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
       });
 
       it("Add filter row", () => {
-        filters[1].Key = "data.surname";
-        filters[1].Type = "String equals";
-        filters[1].Value = "White";
+        filters[1]["Key"] = "data.surname";
+        filters[1]["Type"] = "String equals";
+        filters[1]["Value"] = "White";
 
         cy.ouiaId("add-filter", "PF4/Button").click();
 
         cy.ouiaId("item-1").within(() => {
-          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[1].Key);
-          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[1].Type);
+          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[1]["Key"]);
+          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[1]["Type"]);
         });
         //The filter-value was detached from DOM and we need to find the context again.
         cy.ouiaId("item-1")
           .ouiaId("filter-value", "PF4/TextInput")
-          .type(filters[1].Value);
+          .type(filters[1]["Value"]);
 
         cy.ouiaId("submit", "PF4/Button").should("be.visible").click();
 
@@ -328,20 +328,20 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
       });
 
       it("Add filter row", () => {
-        filters[0].Key = "data.surname";
-        filters[0].Type = "String equals";
-        filters[0].Value = "White";
+        filters[0]["Key"] = "data.surname";
+        filters[0]["Type"] = "String equals";
+        filters[0]["Value"] = "White";
 
         cy.ouiaId("add-filter", "PF4/Button").click();
 
         cy.ouiaId("item-0").within(() => {
-          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[0].Key);
-          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[0].Type);
+          cy.ouiaId("filter-key", "PF4/TextInput").type(filters[0]["Key"]);
+          cy.ouiaId("filter-type", "PF4/FormSelect").select(filters[0]["Type"]);
         });
         //The filter-value was detached from DOM and we need to find the context again.
         cy.ouiaId("item-0")
           .ouiaId("filter-value", "PF4/TextInput")
-          .type(filters[0].Value);
+          .type(filters[0]["Value"]);
 
         cy.ouiaId("submit", "PF4/Button").should("be.visible").click();
 
@@ -488,7 +488,7 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
               );
             });
           filters.forEach((item) => {
-            cy.ouiaId(item.Key, "PF4/TableRow")
+            cy.ouiaId(item["Key"], "PF4/TableRow")
               .find("td")
               .each((cell, cellIndex) => {
                 cy.wrap(cell).should(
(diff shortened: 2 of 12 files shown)


kelvah:
@jomarko FYI there's [also this ticket](https://issues.redhat.com/browse/MGDOBR-1055) covering some of the processors tests

jomarko:
@kelvah @ppadti ok, I am not going to add new tests, just fix existing

jomarko:
I will rebase this PR once the original #229 PR is merged

jomarko:
Decided to rebase PR right now. The changes are not dependent on #229

jomarko:
Unrelated `InstanceCreation` test failed. It is being solved by #226 and #230

```
<testsuites name="Mocha Tests" time="145.3570" tests="4" failures="1">
  <testsuite name="Root Suite" timestamp="2022-10-20T05:34:43" tests="0" file="integration/instance/InstanceCreation.ts" time="0.0000" failures="0">
  </testsuite>
  <testsuite name="the &apos;Create a SE instance&apos; Modal" timestamp="2022-10-20T05:34:43" tests="4" time="145.3570" failures="1">
    <testcase name="the &apos;Create a SE instance&apos; Modal Submit" time="0.0000" classname="Submit">
      <failure message="Timed out retrying after 120000ms: `cy.should()` failed because this element is detached from the DOM.

`&lt;div data-ouia-component-id=&quot;steps-count&quot; data-ouia-component-type=&quot;QE/StackItem&quot; class=&quot;pf-l-stack__item&quot;&gt;1 of 3 ...&lt;/div&gt;`

Cypress requires elements be attached in the DOM to interact with them.

The previous command that ran was:

  &gt; `cy.ouiaId()`

This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add &apos;guards&apos; which delay Cypress from running new commands.

https://on.cypress.io/element-has-detached-from-dom" type="CypressError"><![CDATA[CypressError: Timed out retrying after 120000ms: `cy.should()` failed because this element is detached from the DOM.

`<div data-ouia-component-id="steps-count" data-ouia-component-type="QE/StackItem" class="pf-l-stack__item">1 of 3 ...</div>`

Cypress requires elements be attached in the DOM to interact with them.

The previous command that ran was:

  > `cy.ouiaId()`

This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.
```

cypress/integration/ProcessorTest.ts
```diff
@@ -18,7 +15,14 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
 
       it("Sink processor", () => {
         const processorName: string = "Sink processor";
-        const action = ["Slack", "dev-action", "https://test.app.com/item"];
+        const action = {
```
vpellegrino, line 18: what about having constants used as key in this `action` object?
example:
```
var CONSTANT_A = 0, CONSTANT_B = 1
var state_machine = {
    [CONSTANT_A]: function () {
        return 'a'
    },
    [CONSTANT_B]: function () {
        return 'b'
    }
};
```

in this way you can use them also in lines 72, 84, 91

sjamboro (commented):
IIUC this PR is about removing a flag which causes that several tests were skipped. I can see some refactoring in the action/source data format. It seems that execution of these tests are not change. 

@jomarko thank you for refactoring of this test suite. It is more readable. I have left some comments.

vpellegrino (approved):
Just left one comment about extracting to constants few strings, apart from that it looks good to me

cypress/integration/ProcessorTest.ts
```diff
@@ -18,7 +15,14 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
+          "Action type": "Slack",
+          Channel: "dev-action",
```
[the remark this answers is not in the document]
sjamboro: To be honest ... in future I would like to see here some function which will read information directly from action definition.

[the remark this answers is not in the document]
jomarko: @sjamboro Thank you for this observation. It is related to prettier. https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/232/commits/32886921119f58e878d3907a25a96e2398652cfc

cypress/integration/ProcessorTest.ts
```diff
@@ -406,14 +409,19 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
 
       cy.ouiaId("action-section", "PF4/Text").should("have.text", "Action");
       cy.get("[class='pf-c-description-list__term']")
```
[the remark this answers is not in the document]
jomarko: I sent unification in a separate commit. Lets see CI

cypress/integration/ProcessorTest.ts
vpellegrino: Square brackets are used inside the object definition to have variables/context inside the object's key (https://stackoverflow.com/a/68796045).
Once you have defined a constant for each string, you can use the constant directly in the mentioned rows.

cypress/integration/ProcessorTest.ts
```diff
@@ -18,7 +15,14 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
-        const action = ["Slack", "dev-action", "https://test.app.com/item"];
+        const action = {
+          "Action type": "Slack",
+          Channel: "dev-action",
+          "Webhook URL": "https://test.app.com/item",
+          "Icon Emoji": "Property not configured",
+          "Icon URL": "Property not configured",
+          Username: "Property not configured",
+        };
         const filters = [
```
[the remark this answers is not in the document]
jomarko: looking into it

cypress/integration/ProcessorTest.ts
sjamboro: @jomarko in case that you want to use some constants for Action/Source/Filters labels can you think about importing them from Utils? My future thoughts: it would be nice to import and read these objects: https://github.com/jomarko/sandbox-ui/blob/PR-MGDOBR-1148/mocked-api/schemasData.ts#L240-L290

cypress/integration/ProcessorTest.ts
```diff
@@ -462,22 +473,28 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
     }
 
-    function assertFilters(filters: string[][]) {
-      const filtersHeader: string[] = ["Key", "Type", "Value"];
-
+    function assertFilters(
+      filters: { [key in "Key" | "Type" | "Value"]: string }[]
+    ) {
       cy.ouiaId("filters-section", "PF4/Text").should("have.text", "Filters");
-      if (filters?.length > 0 && filters[0].length > 0) {
```
sjamboro, line 469: Is there any reason why you have removed this condition? Now we are sure that filters has only three types of keys (Key, Type, Value) but someone could input {Key: "data.name", Value: "John"} in such case the code will throw an expection. Is it the expected behaviour? I understand that the previous structure of code is not ideal.

cypress/integration/ProcessorTest.ts
```diff
@@ -310,43 +327,38 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
       });
 
-      onlyOn(bugPresent, () => {
-        it("Add filter row", () => {
-          if (bugPresent) {
-            cy.log(bugPresent);
-            return;
-          }
+      it("Add filter row", () => {
+        filters[0].Key = "data.surname";
```
sjamboro, line 331: @jomarko There could be a problem

cypress/integration/ProcessorTest.ts
```diff
@@ -225,50 +240,50 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
         );
       });
 
-      onlyOn(bugPresent, () => {
-        it("Add filter row", () => {
-          filters[1] = ["data.surname", "String equals", "White"];
+      it("Add filter row", () => {
+        filters[1].Key = "data.surname";
```
sjamboro, line 244: @jomarko there could be a problem

sjamboro (commented):
LGTM ... thank you that you have enabled "add filter row" tests.

Unfortunately, checks are failing so can you take a look at it?
image
```
<testsuite name="Edit Sink Processors" timestamp="2022-10-20T15:20:25" tests="2" time="6.2780" failures="1">
<testcase name="Processor Test Edit Sink Processors Assert and cancel edit form" time="6.2780" classname="Assert and cancel edit form"> </testcase>
<testcase name="Processor Test Edit Sink Processors Add filter row" time="0.0000" classname="Add filter row">
<failure message="Cannot set properties of undefined (setting 'Key')" type="TypeError">
<![CDATA[ TypeError: Cannot set properties of undefined (setting 'Key') at Context.eval (webpack:///./integration/ProcessorTest.ts:244:22) ]]>
</failure>
</testcase>
</testsuite>
<testsuite name="Edit Source Processors" timestamp="2022-10-20T15:20:36" tests="2" time="7.1440" failures="1">
<testcase name="Processor Test Edit Source Processors Assert and cancel edit form" time="7.1440" classname="Assert and cancel edit form"> </testcase>
<testcase name="Processor Test Edit Source Processors Add filter row" time="0.0000" classname="Add filter row">
<failure message="Cannot set properties of undefined (setting 'Key')" type="TypeError">
<![CDATA[ TypeError: Cannot set properties of undefined (setting 'Key') at Context.eval (webpack:///./integration/ProcessorTest.ts:331:22) ]]>
</failure>
</testcase>
</testsuite>
```
It seems that filter structure does not work correctly. Can you take a look at it?

cypress/integration/ProcessorTest.ts
sjamboro: Do you think that there could be: `filters[1]["Key"]`?

cypress/integration/ProcessorTest.ts
sjamboro: Do you think that there could be: `filters[1]["Key"]`?

cypress/integration/ProcessorTest.ts
```diff
@@ -18,10 +15,17 @@ onlyOn(isEnvironmentType(EnvType.Mocked), () => {
...
-          ["data.surname", "String equals", "White"],
+        const action = {
+          "Action type": "Slack",
+          "Channel": "dev-action",
+          "Webhook URL": "https://test.app.com/item",
+          "Icon Emoji": "Property not configured",
+          "Icon URL": "Property not configured",
+          "Username": "Property not configured",
+        };
+        const filters: { [key in "Key" | "Type" | "Value"]: string }[] = [
```
sjamboro, line 26: I think that this definiton could not be interpated as an property of an object (`.Key` notation)

cypress/integration/ProcessorTest.ts
jomarko: I agree this is a place for future improvement. However I do not want to do it as part of this PR, that is already quite large. If I understand correctly the code, even before my refactoring, we didn't validated filter 'deeply'. To me it seems that `input {Key: "data.name", Value: "John"}` was 'valid'  also before my refactoring.

cypress/integration/ProcessorTest.ts
jomarko: I am not `TS` expert programmer, could you please clarify your concern?

is the problem index I used `0` vs `1`  , or `.Key` vs `["Key"]`?

cypress/integration/ProcessorTest.ts
jomarko: similar question as above

cypress/integration/ProcessorTest.ts
jomarko: not sure I understand, could you please try to clarify?

cypress/integration/ProcessorTest.ts
sjamboro: @jomarko  I mean that `.Key` might caused that problem. I am not TS expert too but based on my Java and JS experience:
1.`.Key` reference the property of an object which is saved in the `fitlers` array under index `0`
2. `["Key"]` reference the item in an array/map structure which key is `"Key"` this array/map is saved in the `filters` array under index '1'.

cypress/integration/ProcessorTest.ts
sjamboro: @jomarko  I mean that `.Key` might caused that problem. I am not TS expert too but based on my Java and JS experience:
1.`.Key` reference the property of an object which is saved in the `fitlers` array under index `1`
2. `["Key"]` reference the item in an array/map structure which key is `"Key"` this array/map is saved in the `filters` array under index `1`.

cypress/integration/ProcessorTest.ts
sjamboro: @jomarko I have just highligths the definition of the `filters` variable which is mentioned in the previous notes. I am not sure but this definition says that you can use only `filtres["Key"][1]`, isn't it?

cypress/integration/ProcessorTest.ts
sjamboro: @jomarko the error message is <failure message="Cannot set properties of undefined (setting 'Key')" type="TypeError">

jomarko: ok, pushing separate commit with `["xxx"]` syntax

cypress/integration/ProcessorTest.ts
sjamboro: @jomarko It seems that problems is still here. Do you think that this could help: `filters.push({ Key: "data.name", Type: "String equals", Value: "John" });`?

sjamboro: > @jomarko It seems that problems is still here. Do you think that this could help: `filters.push({ Key: "data.name", Type: "String equals", Value: "John" });`?

The failure was moved. The assertation of the new filter row failed >>> "undefined".

jomarko: @sjamboro done in a separate commit

sjamboro (approved):
LGTM ... thank you ... this PR brings more readable code

Srihari1192 (approved):

.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "trailingComma": "es5",
   "singleQuote": false,
-  "semi": true
+  "semi": true,
+  "quoteProps": "consistent"
```
kelvah, line 5: I was OK with the default configuration, only requiring quotes when needed. But let's give this a try.

kelvah (approved):

ChughShilpa (approved):

jomarko:
Waiting for https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/pull/247

PR #235, opened 2022-10-24
MGDOBR-757: Fix lodash usage as transitive dependency

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-757

### DIFF
--- a/src/app/Processor/ProcessorEdit/useValidateProcessor.ts
+++ b/src/app/Processor/ProcessorEdit/useValidateProcessor.ts
@@ -3,7 +3,7 @@ import {
   ProcessorFormData,
   ProcessorValidation,
 } from "../../../types/Processor";
-import { omit } from "lodash";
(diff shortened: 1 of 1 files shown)


jomarko (commented):
Where can I read sth about this 'dot' syntax please?

vpellegrino (approved):
If I understand correctly, importing the function from the `lodash` container created the transitive dependency issue.
Importing directly from the package declared in `package.json` solves the issue, right?

kelvah:
> Where can I read sth about this 'dot' syntax please?

@jomarko In this case it's just the [name of the package](https://www.npmjs.com/package/lodash.omit) on npm.

> If I understand correctly, importing the function from the `lodash` container created the transitive dependency issue. Importing directly from the package declared in `package.json` solves the issue, right?

Correct.

jomarko (approved):

PR #239, opened 2022-10-26
MGDOBR-1165: Fix heading levels within empty and error UI states

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1165

### DIFF
--- a/src/app/Instance/InstancePage/ProcessorsTabContent.tsx
+++ b/src/app/Instance/InstancePage/ProcessorsTabContent.tsx
@@ -209,7 +209,7 @@ export const ProcessorsTabContent = ({
           >
             <EmptyState variant="large">
               <EmptyStateIcon icon={PlusCircleIcon} />
-              <Title headingLevel="h4" size="lg">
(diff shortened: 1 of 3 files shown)


jomarko (approved):
I think I do not understand properly the rule when `h1` vs `h2` is needed. But basically I agree with increasing the `importance` from `h4`.

I found these 4 places, where different `headingLevel` is used

Screenshot from 2022-10-27 07-52-58

kelvah:
> I think I do not understand properly the rule when `h1` vs `h2` is needed. But basically I agree with increasing the `importance` from `h4`.

@jomarko The general rule is to use headings that reflect the content hierarchy within the page.

For `PageNotFound` and `ErrorDetailWithBoundary`, the error message is the main and only content of the page, so it should be an `h1`.

Within `InstancesListPage` we already have an `h1` for the title of the page. So the "no instances found" message should be the second in importance after the page title, hence I set it to `h2`.

vpellegrino (approved):
Looks good

PR #246, opened 2022-10-31
MGDOBR-1196: "Create bridge" dialog hangs after API unexpected managed errors

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1196

Steps to test:
- run the app against mocked APIs with `npm run start:mocked`
- click on `Create Smart Events Instance`
- enter a name containing `error-test`, like `bridge error-test`
- submit the form by clicking on `Create Smart Events Instance`
- after the API request is done, the dialog exits its loading status and the generic error alert appears on top

### DIFF
--- a/cypress/integration/instance/InstanceCreation.ts
+++ b/cypress/integration/instance/InstanceCreation.ts
@@ -96,6 +96,7 @@ describe("the 'Create a SE instance' Modal", () => {
       });
   });
 
+  // instance with name "error-test" already exists
   it("Submit and expect error", () => {
     const errorInstanceName: string = "error-test";
     cy.ouiaId("create-smart-event-instance", "PF4/Button").click();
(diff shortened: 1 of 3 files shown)


manstis (approved):
LGTM :+1:

jomarko:
Do I understand correctly this is an expected state @kelvah ?
Screenshot from 2022-11-01 08-42-35

kelvah:
> Do I understand correctly this is an expected state @kelvah ? Screenshot from 2022-11-01 08-42-35

@jomarko That's correct. It's an unexpected error that shouldn't happen but we have to cover it anyway.

jomarko (approved):
Fine for me, however I think, we should add a test https://github.com/kelvah/sandbox-ui/pull/7

PR #248, opened 2022-11-01
Fix missing dependencies

jomarko:


jomarko:
it looks like axios 1.1.3 (which the sdk depends upon) has some issues with its exports. It was already reported to them, see https://github.com/axios/axios/issues/5150 and others. They already merged a fix, so it's only a matter of waiting for its next release, plus a new one for sdk too

PR #238, opened 2022-10-25
[no-jira] Fix npm vulnerabilities

jomarko:
npm install was returning:
```
jomarko@localhost sandbox-ui (main)$ npm install

> smartevents-ui@0.0.1 prepare
> husky install

husky - Git hooks installed

added 13 packages, removed 61 packages, changed 5 packages, and audited 1468 packages in 8s

184 packages are looking for funding
  run `npm fund` for details

3 vulnerabilities (2 high, 1 critical)

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
```

This PR contains changes done by `npm audit fix`

PR #251, opened 2022-11-03
NO-JIRA: Fix @xmldom/xmldom security issues

kelvah:
See security alert: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/security/dependabot/6

`@xmldom/xmldom` versions `< 0.7.7` have vulnerability issues. We already fixed this dep issues in the past with `npm audit fix`, moving from `0.7.5` to `0.7.6` (#238).
We would like to avoid doing this for each minor/patch release until we get a good one.

This is the xmldom dependency tree:
```
smartevents-ui@0.0.1 /home/remanuel/workspace/sandbox-ui
└─┬ msw@0.47.4
  └─┬ @mswjs/interceptors@0.17.5
    └── @xmldom/xmldom@0.7.6
```

Unfortunately there's no straightforward way to update xmldom by updating its parent deps. [`@mswjs/interceptors`](https://github.com/mswjs/interceptors/blob/aff787ee28e58737662b31b93e2b7f24faad7beb/package.json#L71) is listing it with `"@xmldom/xmldom": "^0.8.3"`, so our `0.7.6` version is satisfying the dep requirement, running `npm update msw` will do basically nothing.

The only option I had was to manually uninstall and reinstall `msw` to get the latest versions of all its transitive deps.

As a side note, this could be the same reason why renovate is not opening a PR for it. There's no way to fix this by just updating versions inside `package.json` given how the involved transitive deps are declared. (see [`transitiveRemediation` docs](https://docs.renovatebot.com/configuration-options/#transitiveremediation) about remediations not affecting `package.json`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2759,14 +2759,14 @@
       }
     },
     "node_modules/@mswjs/interceptors": {
-      "version": "0.17.5",
-      "resolved": "https://registry.npmjs.org/@mswjs/interceptors/-/interceptors-0.17.5.tgz",
-      "integrity": "sha512-/uZkyPUZMRExZs+DZQVnc+uoDwLfs1gFNvcRY5S3Gu78U+uhovaSEUW3tuyld1e7Oke5Qphfseb8v66V+H1zWQ==",
+      "version": "0.17.6",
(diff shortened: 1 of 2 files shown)


vpellegrino (approved):
Looks good to me

jomarko (approved):
Thank you @kelvah !

PR #253, opened 2022-11-04
MGDOBR-1227: Adopt ResponsiveTable from app-services-ui-components

vpellegrino:
[MGDOBR-1227](https://issues.redhat.com/browse/MGDOBR-1227)


cypress/integration/smoke/BasicTest.ts
```diff
@@ -58,8 +55,7 @@ describe("Basic Elements", () => {
 
   it("Instance header details are available", () => {
     const instanceHeaderDetails = ["Name", "Time created", "Status"];
-    cy.ouiaId("Instances list table", "PF4/Table")
-      .ouiaType("PF4/TableRow")
+    cy.ouiaType("PF4/Table")
```
sjamboro, line 58: We can rewrite this command to use (data-testid and ouiaType)

vpellegrino: Looking at `ResponsiveTable`, it seems that it is not currently supporting `data-testid` for the `table` tag either.

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -196,26 +196,24 @@ const InstancesListPage = (): JSX.Element => {
     resetDeleteInstance();
   }, [resetDeleteInstance]);
 
-  const tableActions = (rowData: TableRow): IAction[] => [
+  const tableActions = (rowData: IRowData): IAction[] => [
```
sjamboro, line 199: https://github.com/patternfly/patternfly-react/blob/main/packages/react-table/src/components/Table/TableTypes.tsx#L90-L92

vpellegrino:
@jomarko @sjamboro, we still have three tests failing: `ProcessorTest.ts`, `instance/InstanceCreation.ts`, and `instance/InstanceEdit.ts`.
I'm struggling to understand why they are failing. Can you please help me to fix them? So that I can officially open this PR for reviews. Thanks

jomarko:
@vpellegrino trying to download and run locally

sjamboro:
@vpellegrino @jomarko It seems that 2 tests `failed because this element is detached from the DOM` and one test failed because `Expected to find element: `[data-ouia-component-type='QE/ResourceStatus'][data-ouia-component-id='ready']`, but never found it`. I need time to review code.

jomarko:
@vpellegrino `ProcessorTest` passed on my machine with these changes https://github.com/vpellegrino/sandbox-ui/pull/6, I am still checking `instance` tests

vpellegrino:
Hi @jomarko,
I'm wondering if there is any update on other test failures.
Appreciation

src/app/Instance/InstancesListPage/InstancesListPage.tsx
```diff
@@ -63,53 +63,56 @@ const InstancesListPage = (): JSX.Element => {
...
+        accessor: "status",
+        label: t("common.status"),
+        formatter: (value: unknown, row?: IRowData): JSX.Element => {
+          const statusString = value;
+          const requestedAt = new Date(
+            (row as BridgeResponse)?.modified_at ??
+              (row as BridgeResponse)?.submitted_at
+          );
+          return (
+            <SEStatusLabel
```
sjamboro, line 105: @vpellegrino how often is updated SEStatusLabel? It seems that when the status is changed than all cells in the row are generated again. So cypress lost reference for the visible element.

I am not sure if it could be caused any callBack. Previously only the value was changed not whole element.

vpellegrino: Columns definitions (so formatters) are already memoized (useMemo above), so I think there is nothing we can do from that point of view.
Is it possible to write tests differently?

sjamboro: I have checked that:
A. this branch has similar Cypress version as main branch => probably no bug in cypress.
B. PF Table changes similar elements and their attributes as ResponsiveTable => See:
image

I am not sure what could cause so many detached DOM element errors. I need to find a way how to rewrite selectors for affected tests. I have tried same logic as @jomarko and it is failing.

```
    cy.ouiaId("Instances list table", "PF4/Table")
      .ouiaId(newInstanceName, "PF4/TableRow")
      .ouiaId("creating", "PF4/Button")
      .should("have.text", "Creating");
    cy.ouiaId("Instances list table", "PF4/Table")
      .ouiaId(newInstanceName, "PF4/TableRow")
      .ouiaId("creating", "PF4/Button")
      .click();
```

sjamboro: @vpellegrino I agree with you. I have checked it as a user and ResponsiveTable is similar to PF4 from a user's point of view. Is possible to use the OUIA standard in selector.

The big disadvantage of this approach is that we lost the reliability of our tests. If I cannot click on the element  by
```
    cy.ouiaId("Instances list table", "PF4/Table")
      .ouiaId(newInstanceName, "PF4/TableRow")
      .ouiaId("creating", "PF4/Button")
      .click();
```
so we need to change the structure of our tests.

vpellegrino: @sjamboro, can you please help me by providing a PR fixing those tests? Thank you very much.

sjamboro: @vpellegrino yes of course. I tried ito fix it yesterday. Unfortumately without any result.

vpellegrino:
@sjamboro @kelvah PR is ready for your review.
Open to suggestions on both tests & production code.

kelvah:
Hi @vpellegrino, quick question.

I see the shared table also supports pagination, although not via `ResponsiveTable` but through `TableView`, which I understand is a wrapper of the first.

Did you already have a look at it? Any thoughts? Could we use it and replace our `TableWithPagination`?

vpellegrino:
> Hi @vpellegrino, quick question.
> 
> I see the shared table also supports pagination, although not via `ResponsiveTable` but through `TableView`, which I understand is a wrapper of the first.
> 
> Did you already have a look at it? Any thoughts? Could we use it and replace our `TableWithPagination`?

Hi @kelvah,

No. As stated in the PR description, the purpose of this task was to use `ResponsiveTable`.
Because it is the component used in https://redhat-developer.github.io/app-services-ui-components/?path=/docs/shared-table-responsivetable--example
image
I think we need to define better the scope of the task.
BTW, if you feel that using `TableView` instead of `TableWithPagination` would benefit our project, I would only ask QE to check if this change would affect the IT suite further.

After briefly looking at `TableView` component definition, there is just one point that comes to my mind.
I see that we need to pass `data` as `null` for showing the loading skeleton.
I'm just wondering if the polling case might be affected by a few issues or if it is managable via the API hook somehow.

kelvah:
> Hi @kelvah,
> 
> No. As stated in the PR description, the purpose of this task was to use `ResponsiveTable`. Because it is the component used in https://redhat-developer.github.io/app-services-ui-components/?path=/docs/shared-table-responsivetable--example. I think we need to define better the scope of the task. BTW, if you feel that using `TableView` instead of `TableWithPagination` would benefit our project, I would only ask QE to check if this change would affect the IT suite further.
> 
> After briefly looking at `TableView` component definition, there is just one point that comes to my mind. I see that we need to pass `data` as `null` for showing the loading skeleton. I'm just wondering if the polling case might be affected by a few issues or if it is managable via the API hook somehow.

@vpellegrino I wasn't necessarily implying that we should adopt `TableView` within this PR/task. Just checking if you already discovered the pagination feature and/or if you had any thoughts about it. Sorry for the confusion.

About the scope of the "shared table" adoption, we can have a quick call to discuss some options.

Looking at the ui-components storybook, I think that we'll have to use `TableView` if we want to also reuse its filtering/searching features at some point. Let's discuss it.

I'll finish my review then. Thank you.

cypress/integration/instance/InstanceCreation.ts
```diff
@@ -44,10 +44,11 @@ describe("the 'Create a SE instance' Modal", () => {
               "This will be ready shortly."
             );
           });
+        cy.wait(150);
```
kelvah, line 47: I've investigated a bit the "detached" issue and I couldn't find other ways to fix it.

Doing two queries with a `0` wait also works, at least locally, but I don't know if it's any better.

```
cy.ouiaId("creating", "PF4/Button").should("be.visible");
cy.wait(0)
cy.ouiaId("creating", "PF4/Button").click();
```

It's clear that the table re-renders shortly after the data is displayed for the first time. You can even see a subtle movement of the columns when it happens. It's likely the table is doing some internal js calculation about the column layout to enable its responsive behavior.

Anyway, I think cypress should at least be able to re-query elements by itself when a detachment occurs. Fortunately it seems they are working on it, they recently merged a fix for this issue, it will be available with cypress 12.

kelvah (approved):
LGTM :+1: 
Nice work!

cypress/integration/instance/InstanceCreation.ts
sjamboro: @kelvah thank you for investigating it. I have thought that the ResponsiveTable must re-render all elements but I did not catch when it happens. The addition of the small wait routine is without problem.

Do you think that this wait can be moved before cy.ouiaId("Instances list table", "PF4/Table")?

vpellegrino: @sjamboro AFAIK, to work, the wait should happen just before the click event.

sjamboro (commented):
Code looks fine. Please let me try it locally. If the table re-renders elements I would like to wait on it before the test starts to assert elements in the table.

By the way, can you rebase this PR? The main contains the fix of SSO cookies form (it should fix problem with UI :: DEV).

cypress/integration/instance/InstanceCreation.ts
sjamboro: @vpellegrino thank you. I thought that the wait is connected with `It's clear that the table re-renders shortly after the data is displayed for the first time.` But probably Cypress require it from any reason.

PR #267, opened 2022-11-24
MGDOBR-1230: Switch to the "v2" TS SDK

ppadti:
JIRA: https://issues.redhat.com/browse/MGDOBR-1230


kelvah:
Hi @vpellegrino, @ppadti, you can start reviewing this PR when you have time.
I've removed all the breaking stuff. I've left some code in the processors area because it wasn't causing any issue. We can eventually decide to keep it or scrap it when we start coding the new processors pages.

Let me know what you think.

@Srihari1192 , @sjamboro, I've skipped all the cypress tests that are not applicable anymore. I don't think we can reuse any of them as they are, so another good option would be to delete them. I'll wait for your opinion on this.

kelvah:
dev tests are failing because the v2 APIs are responding with error 500 at the moment.

kelvah:
> dev tests are failing because the v2 APIs are responding with error 500 at the moment. I'll look into it next week.

Follow-up, V2 APIs are not yet ready on `dev`. We'll have to skip the e2e tests there for now.

Srihari1192:
@kelvah I think its better to skip all the cypress tests for now and keep, it may help for reference during implementation of v2 tests. Once v2 UI e2e tests added will delete all these skipped tests.
@sjamboro wdyt?

sjamboro:
> @kelvah I think its better to skip all the cypress tests for now and keep, it may help for reference during implementation of v2 tests. Once v2 UI e2e tests added will delete all these skipped tests. @sjamboro wdyt?

Hi @kelvah,
I agree with @Srihari1192. It is better to skip tests. The report shows that there are skipped test. So there is a simple reminder that test suite is not ready. We can delete them later.

sjamboro:
The failed test was a problem of [this timeout](https://github.com/ppadti/sandbox-ui/blob/MGDOBR-1230/cypress/integration/instance/InstanceCreation.ts#L84). It is set to 10 secs. I have meassured similar duration localy (9.9)

sjamboro (approved):

vpellegrino (approved):
Looks good to me!

kelvah:
> The failed test was a problem of [this timeout](https://github.com/ppadti/sandbox-ui/blob/MGDOBR-1230/cypress/integration/instance/InstanceCreation.ts#L84). It is set to 10 secs. I have meassured similar duration localy (9.9)

Thanks @sjamboro for looking into it. Did you already create a jira for it? Or do you prefer to fix it here?

sjamboro:
@kelvah I have taken a look at cypress tests. Please let me know when the PR will be finall I will check locally that the feature which is covered by the "flaky" tests is ok.

Srihari1192 (approved):

PR #271, opened 2022-11-30
MGDOBR-1269: Setup Storybook for RHOSE UI

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1269

You can see an example of the storybook created for this PR at the following URL:
https://63863b4610e04c689d81cbcc-mpawqporzz.chromatic.com

I've only added a couple of example stories, it looks a bit empty now. The idea is to add more of them after we merge the v2 PR (after we remove a bunch of stuff).

There's a new github workflow that will build&deploy a new storybook when there are new commits on `main` or inside PRs.
If you scroll inside this PR checks list you'll find a "Storybook Publish" entry. Clicking on the correspondent "Details" link will open the latest storybook built from the last commit of the PR.


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          fetch-depth: 0
+      - name: Install dependencies
+        run: npm ci
+        #👇 Adds Chromatic as a step in the workflow
+      - uses: chromaui/action@v1
+        # Options required for Chromatic's GitHub Action
+        with:
+          #👇 Chromatic projectToken, see https://storybook.js.org/tutorials/intro-to-storybook/react/en/deploy/ to obtain it
+          projectToken: 65b519905a4d
```
kelvah, line 25: Leaving the chromatic project token as it is here because otherwise it wouldn't be possible to publish storybooks for external forks PRs. [More info about it](https://www.chromatic.com/docs/github-actions#run-chromatic-on-external-forks-of-open-source-projects).
Btw `app-services-ui-components` [does the same](https://github.com/redhat-developer/app-services-ui-components/blob/5a321948228d61253019065c0c8eae042bc54f36/.github/workflows/reusable-build.yml#L95).

.storybook/manager-head.html
```diff
@@ -0,0 +1 @@
+<meta name="robots" content="noindex" />
```
kelvah, line 1: Preventing search engines from crawling and indexing.

.storybook/preview-head.html
```diff
@@ -0,0 +1,5 @@
+<style>
+  #root {
+    height: 100%;
+  }
+</style>
```
kelvah, line 5: Setting the height of the components preview container to an explicit value. This is useful for components that rely to a parent element with an explicit height.

package.json
```diff
@@ -21,12 +21,16 @@
...
+    "generate-error-catalog": "node scripts/error-catalog-sdk/generate_error_catalog.mjs",
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook",
+    "chromatic": "chromatic"
   },
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
     "@patternfly/patternfly": "4.219.2",
+    "@patternfly/react-charts": "6.94.7",
```
kelvah, line 33: Besides all the storybook and chromatic related dependencies, I had to add this one too.
It's a peer dependency of `app-services-ui-components` and without it I was having issues while rendering our components within storybook.

tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "mocked-api",
     "setupTests.ts"
   ],
+  "exclude": ["dist", "node_modules"],
```
kelvah, line 12: `stories` files are already correctly excluded [in the build ts configuration](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox-ui/blob/5c33d7cd8c7315ac2873314a5cf220776fba4f5d/tsconfig.build.json#L20), but they shouldn't be in the "dev" tsconfig, hence the override here.

vpellegrino (commented):
Great job @kelvah 
I would just add a new section to `README.md` about storybook usage and the newly added commands.

kelvah:
> Great job @kelvah I would just add a new section to `README.md` about storybook usage and the newly added commands.

@vpellegrino done. 
I've only added the command to run it locally because the deployment is automated.
We could also add the public link to the one from `main` as soon as we have some more stuff to share.

vpellegrino (approved):
Thank you @kelvah

sjamboro (approved):
Thank you @kelvah

PR #262, opened 2022-11-18
MGDOBR-1041: Check Environment Configuration in Test Suite

sjamboro:
https://issues.redhat.com/browse/MGDOBR-1041

PR #273, opened 2022-12-02
MGDOBR-1041: fix errors in Typescript build

sjamboro:
PR #262 contains several errors in Typescript build. It seems that is not possible to import trivialy objects into command files.


kelvah (approved):
LGTM. Thanks @sjamboro.

PR #279, opened 2022-12-15
MGDOBR-1229: Add support for filter in our overview pages

vpellegrino:
**Jira:** https://issues.redhat.com/browse/MGDOBR-1229

image

A few points:

1. Query params are preserved when going back with the browser's button but not when clicking on a link in the application. What should we do about it? What is the expected user experience?
2. I created a hook for managing page parameters for pages using table overviews to favor reusability. We can make this hook configurable (for instance, I may want to add the `name` filter and not `status`). Just let me know.
3. I just added a few statuses supported by the API. Should we add all of them? Support any other not in the current list?
image
4. It seems we have a defect (in `app-services-ui-component`) related to multiple chips when removing one of them:
multiple_chips_filter_defect

### DIFF
--- a/locales/en/smart-events.json
+++ b/locales/en/smart-events.json
@@ -41,8 +41,7 @@
       "ready": "Ready",
       "creating": "Creating",
       "deleting": "Deleting",
-      "failed": "Failed",
-      "preparing": "Preparing"
+      "failed": "Failed"
     },
     "resourceSteps": {
       "pending": {
--- a/locales/en/smart-events.json
+++ b/locales/en/smart-events.json
@@ -3,6 +3,7 @@
     "access": "Access",
     "actions": "Actions",
     "addressFormErrors": "Address form errors to proceed.",
+    "adjustYourFilters": "Adjust your filters and try again, or <0>clear all filters</0>.",
     "ago": "ago",
     "cancel": "Cancel",
     "close": "Close",
@@ -40,7 +41,8 @@
       "ready": "Ready",
       "creating": "Creating",
       "deleting": "Deleting",
-      "failed": "Failed"
+      "failed": "Failed",
+      "preparing": "Preparing"
     },
     "resourceSteps": {
       "pending": {
--- a/mocked-api/handlers.ts
+++ b/mocked-api/handlers.ts
@@ -15,6 +15,7 @@ import { schemaCatalogData, schemasData } from "./schemasData";
 import omit from "lodash.omit";
 import cloneDeep from "lodash.clonedeep";
 import { cloudProvidersData, cloudRegions } from "./cloudProvidersData";
+import { QueryOptions, QuerySelector } from "@mswjs/data/lib/query/queryTypes";
 
 // api url
 const apiUrl = `${process.env.BASE_URL ?? ""}${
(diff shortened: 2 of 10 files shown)


kelvah:
Hi @vpellegrino, answering inline.


> [...]
> A few points:
> 
> 1. Query params are preserved when going back with the browser's button but not when clicking on a link in the application. What should we do about it? What is the expected user experience?

The current behavior is totally fine. We expect to be able to recover previous searches/filters only via browser history for now.

> 2. I created a hook for managing page parameters for pages using table overviews to favor reusability. We can make this hook configurable (for instance, I may want to add the `name` filter and not `status`). Just let me know.

If it's not a big deal I would make it configurable. It's likely there will be some differences between bridges and processors/connectors at some point.

> 3. I just added a few statuses supported by the API. Should we add all of them? Support any other not in the current list?

I think we should use the same "user-facing" statuses we have in the status label component, so `ready`, `creating`, `failed`, `deleting`. For example setting the filter to `creating` will show all the resources in one of the creating statuses (`accepted`, 'preparing` and `provisioning`). Does it make sense? it's the same behavior adopted by kafka.

> 4. It seems we have a defect (in `app-services-ui-component`) related to multiple chips when removing one of them:

It looks like a bug from here too. I wouldn't consider it a blocker for this PR. Let's find out what's the best place to report it.

manstis:
Hey @kelvah 

> I think we should use the same "user-facing" statuses we have in the status label...

I think `v2` is a good opportunity to do https://issues.redhat.com/browse/MGDOBR-983.

Is it something you'd still like? (Only in `v2`).

kelvah (commented):
I've found a small issue in the bridge overview (the same could happen for processors but I didn't check).

If a set a status filter (like `failed`) and then I create a new bridge, the table will reload all the items without search filters. The filters will reappear with the first polling request after the creation of the bridge.

vpellegrino:
@kelvah I worked on user-facing statuses + fixed the defect you found (thanks for catching it).
On Monday I will try to work on the last point (make hook configurable if possible)

vpellegrino:
@kelvah, @ppadti PR si ready to be reviewed

kelvah:
Hi @vpellegrino, I've spotted another minor issue.
When applying a filter, the total number of rows and the pagination info don't get updated accordingly to the results.
My guess is it's only related to our mocked APIs needing some tuning.

vpellegrino:
> Hi @vpellegrino, I've spotted another minor issue. When applying a filter, the total number of rows and the pagination info don't get updated according to the results. My guess is it's only related to our mocked APIs needing some tuning.

@kelvah yeah, it's related to the mocked APIs.
This is basically due to the fact that data fetching and data count are two distinct operations. Let me take a few time to see if I can improve that part

locales/en/smart-events.json
```diff
@@ -40,7 +41,8 @@
       "ready": "Ready",
       "creating": "Creating",
       "deleting": "Deleting",
-      "failed": "Failed"
+      "failed": "Failed",
+      "preparing": "Preparing"
```
kelvah, line 45: Why did you add the `preparing` translation here? :thinking: 
I couldn't find any usages.

kelvah (approved):
LGTM :+1: 
Thanks @vpellegrino

locales/en/smart-events.json
vpellegrino: Probably, at some point, I was introducing it for the filter. BTW, removing it!

ppadti (approved):

PR #286, opened 2023-01-04
NO-JIRA: Fix Trans component usage 

kelvah:
`EmptyState` stories inside storybook have some missing translations.
According to `react-i18next` [docs](https://react.i18next.com/latest/trans-component), the `Trans` component does not re-render when the dictionary is loaded/changed. This is an issue when running components in isolation without any `useTranslation` usage triggering re-renders.

It should be fixed now.


vpellegrino (approved):
Looks good!

PR #289, opened 2023-01-10
MGDOBR-1339: "Create processor" header component poc

kelvah:
JIRA: https://issues.redhat.com/browse/MGDOBR-1339

Initial implementation of the processor editing header.

PF doesn't provide a react component for inline editing, so I've used [their html code only](https://www.patternfly.org/v4/components/inline-edit).

I didn't add the actions dropdown because it's likely actions will not be available while creating/editing a processor. We can come back to this later if necessary.

### DIFF
--- a/src/app/components/POCs/ProcessorEditHeader/ProcessorEditHeader.stories.tsx
+++ b/src/app/components/POCs/ProcessorEditHeader/ProcessorEditHeader.stories.tsx
@@ -1,0 +1,41 @@
+import React, { useState } from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import ProcessorEditHeader from "@app/components/POCs/ProcessorEditHeader/ProcessorEditHeader";
+
+export default {
+  title: "PoCs/Processor Edit Header",
+  component: ProcessorEditHeader,
+  args: {
+    onCreate: () => {},
+    showCreateAction: true,
+  },
+} as ComponentMeta<typeof ProcessorEditHeader>;
+
+const Template: ComponentStory<typeof ProcessorEditHeader> = (args) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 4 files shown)


vpellegrino (approved):
Looks good, thank you @kelvah

src/app/components/POCs/ProcessorEditHeader/ProcessorEditHeader.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const Template: ComponentStory<typeof ProcessorEditHeader> = (args) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { onNameChange, name, ...rest } = args;
+  const [value, setValue] = useState(args.name);
+
+  return (
+    <div style={{ padding: "3em 0" }}>
+      <ProcessorEditHeader
+        onNameChange={(newValue): void => setValue(newValue)}
```
sjamboro, line 22: @kelvah Please what is the purpose of this code? Does it change the 'name' attribute when I set it in story book UI?

sjamboro (approved):
Thank you @kelvah ... LGTM

src/app/components/POCs/ProcessorEditHeader/ProcessorEditHeader.stories.tsx
kelvah: Hi @sjamboro, it's meant to demonstrate the component behavior when trying to set a new name.

`ProcessorEditHeader` keeps an internal state for the name value only to manage the editing phase, meaning controlling the input field and allowing canceling changes.

After renaming the processor and confirming the changes by clicking on the tick icon, the new value is passed up to its parent via `onNameChange`, where the actual state for the name value is stored.

That's the reason why some state management is needed inside the story, otherwise confirming the changes would have no visible effect inside storybook.

PR #304, opened 2023-01-30
MGDOBR-1362: Add new bridge overview to RHOSE UI

ppadti:
JIRA: https://issues.redhat.com/browse/MGDOBR-1362


src/app/Instance/BridgeOverview/BridgeOverview.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+      processorList: processorData,
+    });
+
+    await waitForI18n(comp);
+
+    expect(comp.queryByText("No processors")).not.toBeInTheDocument();
+    expect(comp.queryByText("Processor one")).toBeInTheDocument();
+    expect(comp.queryByText("Processor two")).toBeInTheDocument();
+    expect(comp.queryByText("Processor three")).toBeInTheDocument();
+    expect(comp.queryByText("Processor four")).toBeInTheDocument();
```
kelvah, line 83: You can loop through the data instead of manually check each one:

```suggestion
    processorData.map((processor) => {
      expect(comp.queryByText(processor.name)).toBeInTheDocument();
    });
```

You can eventually do the same to check the href's of all the links.

kelvah (commented):
Nice work @ppadti!
I've left only minor observations.

src/app/Instance/BridgeOverview/Components/BOGettingStarted.tsx
```diff
@@ -16,17 +16,18 @@ import {
...
     <Card isExpanded={isCardExpanded}>
       <CardHeader
+        toggleButtonProps={{ "aria-label": "getting started toggler" }}
         onExpand={(): void => {
           setIsCardExpanded(!isCardExpanded);
         }}
       >
         {isCardExpanded ? (
-          <CardTitle>Getting Started</CardTitle>
+          <CardTitle>Getting started</CardTitle>
```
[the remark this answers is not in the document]
ppadti: As we decided not to add the translation for getting started page for now. Should I add translation only for the title?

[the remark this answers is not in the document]
kelvah: No, you're right. I've forgot it was the getting started component. Ignore my comment.

src/app/Instance/BridgeOverview/BOStoriesHelper.ts
```diff
@@ -12,6 +12,7 @@ export const processorData = [
     status: ManagedResourceStatus.Ready,
     flows: [],
     owner: "",
+    url: `/instance/${instanceId}/processor/${"a72fb8e7-162b-4ae8-9672-f9f5b86fb3d7"}`,
```
kelvah, line 15: I don't particularly like this approach. `url` is not a property of `ProcessorResponse`, so adding it here could be misleading.

src/app/Instance/BridgeOverview/BridgeOverview.test.tsx
```diff
@@ -77,15 +77,14 @@ describe("Bridge Overview", () => {
...
-      "href",
-      "/instance/3543edaa-1851-4ad7-96be-ebde7d20d717/processor/a72fb8e7-162b-4ae8-9672-f9f5b86fb3d7"
-    );
+
+    processorData.map((processor) => {
+      expect(comp.queryByText(processor.name)).toBeInTheDocument();
+      expect(comp.getByRole("link", { name: processor.name })).toHaveAttribute(
+        "href",
+        processor.url
+      );
```
kelvah, line 86: Following my previous comment, I would prefer to check against a url you compose directly here, instead of retrieving it from the processor list. Just to make things more clear.

More or less like this:
 
```suggestion
      expect(comp.getByRole("link", { name: processor.name })).toHaveAttribute(
        "href",
        `/instance/${instanceId}/processor/${processor.id}"`
      );
```

Wdyt?

kelvah (approved):

src/app/Instance/BridgeOverview/BridgeOverview.test.tsx
ppadti: Yes. This gives more clarity. Thank you for the suggestion. I will update it.

src/app/Instance/BridgeOverview/Components/BODashboardTableView.tsx
```diff
@@ -47,69 +62,76 @@ export const BODashboardTableView = (
...
+                                className={
+                                  "BO-Dashboard__resource__truncated-string"
+                                }
+                              />
+                            </Link>
+                          </Text>
+                        </TextContent>
+                      </FlexItem>
+                      <FlexItem>
+                        {item.status != ManagedResourceStatus.Ready && (
```
sjamboro, line 99: IIUC we do not want to show the ready status for each processor.

src/app/Instance/BridgeOverview/Components/BOProcessorList.tsx
```diff
@@ -1,35 +1,88 @@
...
+                }}
+                variant={"secondary"}
+              />
+            ) : (
+              <BODashboardTableView
+                name={t("common.processors")}
+                createButton={{
+                  title: t("processor.createProcessor"),
+                  onCreate: onCreateProcessor,
+                  isDisabled: bridgeStatus !== ManagedResourceStatus.Ready,
```
sjamboro, line 78: Does we have state of BODashboardTableview in storybook when the bridge status is not ready?

sjamboro (approved):
LGTM ... thank you @ppadti 

I left just minor comment.

src/app/Instance/BridgeOverview/Components/BODashboardTableView.tsx
ppadti: Yes. We don't display "Ready" status for the processors.

src/app/Instance/BridgeOverview/Components/BOProcessorList.tsx
ppadti: We don't have story for the Tableview when the bridge is failed, but we have it for the "BOEmptystate".

Srihari1192 (approved):

src/app/Instance/BridgeOverview/Components/BOProcessorList.tsx
sjamboro: Yes, but it is the different usecase. IIUC this code is about accesibility of actions / buttons when the user has the link (url) but the bridge is not ready (for example failed). I can set this state manually in the story book. I just want to be sure that we do not want to catch this state in our storybook.

PR #365, opened 2023-03-28
chore: Namespace Selection

Kells512:


kelvah (commented):
Hi @Kells512!

I have some minor comments before merging this one.

Could you please group all the Namespace variants in a single folder? 
You could also move it inside `src/app/components/POCs`, I think.
Just to keep things more organized. 

Also I've noticed that you added a git submodule in a new `sandbox-ui` empty folder. I guess it wasn't intentional. If so, could you get rid of it?

PR #376, opened 2023-04-17
Topology view component PoC

ppadti:


package.json
```diff
@@ -36,6 +36,7 @@
     "@patternfly/react-styles": "4.92.6",
     "@patternfly/react-table": "4.113.0",
     "@rhoas/app-services-ui-components": "2.20.9",
+    "@patternfly/react-topology": "^4.91.27",
```
kelvah, line 39: Could you switch to the latest version of this package? Also with an exact number, no ranges.

webpack.common.js
```diff
@@ -13,7 +13,10 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const MonacoWebpackPlugin = require("monaco-editor-webpack-plugin");
 
 const isPatternflyStyles = (stylesheet) =>
-  stylesheet.includes("@patternfly/react-styles/css/") ||
+  (stylesheet.includes("@patternfly/react-styles/css/") &&
+    !stylesheet.includes(
+      "@patternfly/react-styles/css/components/Topology/"
+    )) ||
```
kelvah, line 19: Could you please add a comment explaining why we needed this?
We'll likely forget it.
The issue here was that the styles of the topology component are not included in the patternfly-base stylesheet. We needed to allow them to be imported from the `@patternlfy/react-styles` repo, hence the change.

kelvah (commented):
Thanks @ppadti! It looks good!

Also, after removing the green "success" style from regular nodes, the whole diagram looks a bit too grayish now. What you think if we apply the blue border to all nodes? Is it feasible with little effort?

package.json
kelvah: Nitpicking, I would also move it up to keep the alphabetical order of packages.

ppadti:
> Also, after removing the green "success" style from regular nodes, the whole diagram looks a bit too grayish now. What you think if we apply the blue border to all nodes? Is it feasible with little effort?

Yes. I can do that.

kelvah:
@ppadti Let me know if/when the PR is ready for review.

kelvah (approved):
LGTM

src/app/components/POCs/TopologyView/TopologyView.tsx
```diff
@@ -0,0 +1,257 @@
...
+  LayoutFactory,
+  Model,
+  Node,
+  EdgeModel,
+  NodeModel,
+} from "@patternfly/react-topology";
+
+import NodeInformation from "./NodeSideBarDetails";
+
+export interface TopologyViewDemoProps {
```
kelvah, line 45:
```suggestion
export interface TopologyViewProps {
```

